PR #8, opened 2021-02-19
Fix Alignment

ncitron:
The header and wrapper-main elements were misaligned so that they lean to the left on wider monitors.


adamavenir:
hey thanks for this @ncitron. appreciate it.

PR #15, opened 2021-03-03
add timer, current contest

adamavenir:


adamavenir:
Keeping it simple for now while getting started working on the new rev of the site. 

Also made it so the timer doesn't display a negative countdown when we pass the time threshhold ;)

Screen Shot 2021-03-03 at 11 06 40 AM

zscole (approved):

zscole:
just merged, but maybe we should link to their twitter

adamavenir:
@zscole do you mean on the logo, too? I'm currently linking the headline text to their twitter account :)

Screen Shot 2021-03-03 at 11 45 10 AM

PR #18, opened 2021-03-03
Rebuild site on gatsby/react

adamavenir:
This will need DNS config before it can go live.


zscole (commented):
what does DNS need to be updated to? it will hosted on netflify, right?

PR #30, opened 2021-05-12
Add reporter form

dstaley:
This PR migrates the reporting form from contest-specific repos into one generic submission form. It dynamically sets data based on the currently running contests.

Additionally, this PR switches the repo to use Yarn v1, since Gatsby currently doesn't play nice with npm@v7.


adamavenir:
@dstaley omgggggg u r grt

adamavenir:
This is really great, @dstaley. A few requests:

1. Can you use the format `YYYY-MM-slug` for the id (where MM is the month of the start date of the contest), as it's what we're using for the contest repos in this org. Then everything just matches :) Maybe we even make that a 'thing' as part of the contest json and use it to derive other urls.

2. When someone submits the form and goes to add another, it would be great if the handle would pre-populate alongside the other pre-filled fields.

3. We should have a link under the listing that says 'Don't see your name? Add your handle here to register' and link to the code423n4.com github repo docs.

PR #32, opened 2021-05-14
Updating /reports to be more responsive

DianaPerkinsDesign:
I centered the tiles, and made them a bit squishier on mobile. 

I think the background issue on mobile is Chrome's way of dealing with a horizontal scrollbar in the dev tools, so making the tiles more of an auto width fixed it on that page. There's more improvements I could make for mobile here, but it'd take more time to think through. 

For the background issue on the individual report pages, same thing, and I think that one's caused by some very long words that the browser isn't cutting. We could target them and force them to break with a `word-break: break-all;` maybe?

### DIFF
--- a/src/styles/global.css
+++ b/src/styles/global.css
@@ -92,7 +92,12 @@ section:first-of-type {
 
 .wrapper-report {
   display: block;
+  max-width: 900px;
   margin: 0 auto;
+  display: grid;
+  grid-template-columns: repeat(auto-fill, minmax(390px, 1fr));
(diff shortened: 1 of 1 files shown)


adamavenir:
@DianaPerkinsDesign Thanks!

>  I think that one's caused by some very long words that the browser isn't cutting. We could target them and force them to break with a word-break: break-all; maybe?
Good catch. Are they inside <code> blocks perchance? Cos maybe we just apply that style to those.

PR #35, opened 2021-05-18
Merge handles directories

### LINKED ISSUE #34: Fix warden handle data location, docs, identify any missing
Because of outdated docs, we have some wardens not showing up on the leaderboard who have findings in the data set.

The contents of the `data/handles folder simply needs to get moved from `_data/handles` instead. It would be ideal to double check that these handles match the handles in `_data/findings/findings.json`.

And we should update the docs, too, as @alexon1234 pointed out in #33:

> If [json files] should be added inside _data, we should update the README.md because in there it says
> `1. Add a JSON file for yourself at data/handles`

- [ ] update readme for correct path to data
- [ ] relocate data folder to _data
- [ ] check that warden handles in `_data/findings/findings.json` match the handles added to `_data/handles`
- [ ] identify any handles that are in `findings.json` but aren't yet on the leaderboard and create an issue with any wardens who haven't added data to handles folder yet.

dstaley:
This PR merges in the `data/handles` content with the `_data/handles` directory, and updates the README to reflect that.

Fixes #34


adamavenir:
looks great. thank you, @dstaley!

PR #37, opened 2021-05-19
Add `findingsRepo` field to contests

dstaley:
This adds a new `findingsRepo` field to contests. I know typically they're in the same format as the normal repo, but I don't want to leave that up to chance.


adamavenir (commented):
@dstaley seriously appreciate your attention to detail here. thank you for this addition.

PR #40, opened 2021-05-20
Added 0xsomeone to handles

alex-ppg:
As title says, added "0xsomeone" handle info and image.


dstaley:
Thank you! You're on the leaderboard now :D

PR #43, opened 2021-05-21
Add shw handle

x9453:
As title, add shw's handle and avatar.


_data/handles/shw.json
```diff
@@ -0,0 +1,5 @@
+{
+  "handle": "shw",
+  "image": "./avatars/shw.jpg",
```
dstaley, line 3: Should be `./avatars/shw.jpeg`.

dstaley (changes requested):

_data/handles/shw.json
x9453: Fixed, thanks for correcting!

dstaley:
Woohoo! Thank you! You're in the third slot on the leaderboard now :D

PR #71, opened 2021-06-14
adding gro to site

itsmetechjay:


ninek9:
@itsmetechjay we also need the contests.csv updated in the _data/contests folder iirc

PR #90, opened 2021-06-22
Add Sparta Wardens

SamusElderg:
Add Sparta Wardens


ninek9 (changes requested):
Welcome! The image file type for SamusElderg needs to match the image file, so make the appropriate change and we can merge it.

SamusElderg:
> Welcome! The image file type for SamusElderg needs to match the image file, so make the appropriate change and we can merge it.

Whoops rookie mistake sorry about that!
Fixed now.
Thanks!

ninek9 (approved):

PR #99, opened 2021-06-24
fixed three errors, still notice the website doesn't render backticks

loudoguno:


ninek9 (approved):

adamavenir:
Can you give an example of what you're seeing? The backtick markdown seems to be rendering as expected to me.

adamavenir:
Oh I see you did leave an example here https://github.com/code-423n4/code423n4.com/issues/104

PR #101, opened 2021-06-24
on NFTX, fixed some typos, corrected judges name and a few other things

loudoguno:


_reports/2021-05-nftx.md
```diff
@@ -88,7 +88,7 @@ Recommend changing to: `bool tokensReceived = returnData.length == 1 && abi.deco
 
 **[0xKiwi (NFTX) confirmed](https://github.com/code-423n4/2021-05-nftx-findings/issues/46):**
 
-**[cemozerr (Judge) commented](https://github.com/code-423n4/2021-05-nftx-findings/issues/46#issuecomment-848259748):**
```
ninek9, line 91: This is his Github handle so it's okay to leave it that way in the findings.

ninek9 (changes requested):
See my note inline regarding the judge name.

PR #149, opened 2021-07-21
Update contests.csv

itsmetechjay:
Add Notional and Hide Spartan from contests page.


ninek9 (changes requested):
sorry i didn't communicate or update Notion - the dates are 8/26 - 9/8

itsmetechjay:
Ok I think I made that change and it created another commit.  Is that right?

ninek9 (approved):

PR #156, opened 2021-07-24
Remove .reverse() and change to ASC time sort

bighitbiker3:
# Overview

I'm pretty sure this was a cache issue. The `DESC` order was actually _not_ the order we wanted, which is why we'd reverse it. But upon a page change it'd grab the cached, reversed `DESC` data and then reverse it again.

This just grabs the filtered, `ASC` ordered data.

### DIFF
--- a/src/components/ContestList.js
+++ b/src/components/ContestList.js
@@ -4,7 +4,7 @@ import Contest from "./Contest";
 const ContestList = ({ contests }) => {
   return (
     <>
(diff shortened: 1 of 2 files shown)


dstaley (approved):
man gatsby is so freakin' weird sometimes. I wouldn't expect anything other than the GraphQL query to be cached in this case. That being said, I think it's much better to perform the sort in the query rather than at render, so thank you for this!

PR #157, opened 2021-07-27
Add TweetEmbed component

dstaley:
Usage:

```tsx
<TweetEmbed tweetId="1400466820759375876" />
```


ninek9:
it looks ok to me but i don't really know for sure. not sure who to have check these, maybe @adamavenir in the absence of another react dev. maybe @bighitbiker3 would be willing to give a look?

bighitbiker3:
@ninek9 No prob! Looking now

src/components/TweetEmbed.js
```diff
@@ -0,0 +1,42 @@
...
+        </div>
+    );
+};
+
+const TweetEmbed = ({ tweetId }) => {
+    const [isInClient, setIsInClient] = React.useState(false);
+
+    useEffect(() => {
+        setIsInClient(true);
+    }, []);
```
bighitbiker3, line 37: Is the purpose of this to not jar the UI while the `Tweet` component loads the tweet? If not, what's it for?

dstaley: Gatsby won't run `useEffect` hooks during SSR, so this allows us to statically render a placeholder component for the initial response, and then have the Twitter embed code take over once it's rendered on the client. @DianaPerkinsDesign is working on a separate PR based on this one that will render a bunch of tweets, so this might all make a lot more sense once her PR is in!

PR #159, opened 2021-07-28
Create xfiles.json

Felixjosemon:
add team xfiles as warden


ninek9 (changes requested):
for teams we also need the individuals' handles registered, so you'll need to add those before this will validate.

Felixjosemon:
@ninek9 done!

ninek9 (approved):

PR #169, opened 2021-08-05
Fixed formatting in judging-criteria.md

robriks:
Line 94 displayed duplicate text from lines 15-17, rather than iterating the highest scale of resource intensity required for the hack. 

Duplicate text in question ironically referred to bounty sharing between duplicate bug submissions by Wardens. If Warden submissions are graded by formatting, how much is my bounty for fixing a bug off-chain + on code423n4 website? ;)


DianaPerkinsDesign:
Great find! 🙌  We moved this info to docs.code423n4.com a few months ago, and it seems we missed deleting this file (which also has wrong info on it 😄) . Thanks! I'll make a new PR to remove it entirely.

PR #172, opened 2021-08-05
Removing old content files

DianaPerkinsDesign:
Got approval from 🧦  to remove old content files that weren't displayed on the site. 

Question for you @dstaley : I was going to delete the directory also and the stuff that correlates in the config and readme, but I ended up not doing that in case we want to use it in the future. What do you think?


dstaley:
You'll need to remove [these lines](https://github.com/code-423n4/code423n4.com/blob/main/gatsby-config.js#L46-L52) as well.

I'm cool with removing the pages stuff from the README since we're not using it anymore.

PR #191, opened 2021-08-18
Adding message bar

DianaPerkinsDesign:
Adding message bar - Cosmos flavored to start. 

This also changes how the header works: 
- header now lives outside of `<main>`
- `<header>` is now inside header component
- navigation is flexbox


adamavenir (approved):
thank youuuu @DianaPerkinsDesign

PR #228, opened 2021-08-30
adds 0xean

0xean:
adds 0xean as a warden


ninek9 (changes requested):
just one tweak, need to match the filename between the file and the json value.

0xean:
good catch, thanks!

ninek9 (approved):

PR #230, opened 2021-08-31
added visor, gro, and connexts final reports for publishing

loudoguno:
doing all 3 reports ready for publishing in bulk


ninek9 (approved):

ninek9 (changes requested):
looks like we have to capitalize "Visor" in the report header for it to match their org json file

ninek9 (approved):

PR #256, opened 2021-09-14
Update Sushi contests and add PoolTogether

itsmetechjay:


adamavenir (commented):
let's phrase it "Sushi Trident contest phase 1" and "Sushi Trident contest phase 2"

adamavenir (approved):

adamavenir:
@dstaley can you take a look at why this is failing?

dstaley:
@adamavenir @itsmetechjay Looks like `Sushi Trident contest phase 2` is using the same contest id (29) as phase 1. It needs to be a new ID. Looks like it should be 35.

Additionally, it looks like our validate script has an error, so that's why the log wasn't helpful for this case. I'll open a PR to fix that in a moment :)

PR #264, opened 2021-09-16
add broccoli team

Jonah246:


dstaley (changes requested):
Hi @Jonah246! Thank you for registering your team. While using an emoji as your team name _technically_ works, we can't guarantee it will _always_ work, and we also can't promise to fix issues that arise from its usage. Would you mind changing your name to an alphanumeric version instead? You can also use hyphens and underscores. Thank you for understanding!

Jonah246:
> Hi @Jonah246! Thank you for registering your team. While using an emoji as your team name _technically_ works, we can't guarantee it will _always_ work, and we also can't promise to fix issues that arise from its usage. Would you mind changing your name to an alphanumeric version instead? You can also use hyphens and underscores. Thank you for understanding!

Just changed the name.
Thanks for your replying and all the hard works you've done in making this works. Really appreciate it.

ninek9 (approved):

PR #272, opened 2021-09-17
Team GreyArt

STYJ:
Hi, I'm removing my old smurf account (0xSinkie, never used) and shared account (Greiart shared w/ hickuphh3) because I've decided to do this with my real identity instead.

My teammate will still using his anon account because he prefers to stay that way.

Thanks


dstaley:
@STYJ Can you undo the delete of the `greiart` account? It's [attached to several findings](https://github.com/code-423n4/code423n4.com/blob/main/_data/findings/findings.csv#L1491), and we don't allow the removal of accounts that have submitted a finding.

STYJ:
> @STYJ Can you undo the delete of the `greiart` account? It's [attached to several findings](https://github.com/code-423n4/code423n4.com/blob/main/_data/findings/findings.csv#L1491), and we don't allow the removal of accounts that have submitted a finding.

Done! Please lmk what else you need.

STYJ:
Hey guys, can I get an update on this?

ninek9 (approved):

PR #278, opened 2021-09-17
Update warden registration instructions

dstaley:
This PR updates our README with instructions on how to register, including a note about making sure their Discord name matches their handle.


itsmetechjay (approved):
This looks great @dstaley. I approve!  One Q though: Who gets notified about the warden registrations?  or is it just a matter of checking?

dstaley:
It'd be the same notifications as currently (which I'm not aware of any). Once we add auto role assignment we can probably pipe new registrations into Discord in one of our channels.

PR #300, opened 2021-09-23
Improve mobile view for reports list

dstaley:
This PR creates a mobile layout for `/reports` that switches to the desktop layout at `>=1024px`.


src/components/ReportList.js
```diff
@@ -1,10 +1,16 @@
...
+  return (
+    b.node.frontmatter.contest.contestid - a.node.frontmatter.contest.contestid
+  );
+}
+
 const ReportList = ({ reports }) => {
   return (
     <>
-      {reports.reverse().map((report) => (
+      {reports.sort(sortByContestId).map((report) => (
```
DianaPerkinsDesign, line 13: Can we sort by start date instead? I imagine they're not in perfect order when using id since they could be created in a different order than they ran.

DianaPerkinsDesign (commented):
Styles look great! I left a question in `ReportList.js`.

DianaPerkinsDesign (approved):

PR #305, opened 2021-09-29
Updated user handle

0xleastwood:


itsmetechjay (approved):

ninek9:
@dstaley do we have a way of changing handles?

ninek9:
@0xleastwood i chatted with @dstaley offline - at this time we only have a manual process for updating these due to previous findings being associated with the previous handle. we've added the ability to edit handles as an enhancement in our backlog.

0xleastwood:
Just realised I never got back on this. All good! I imagined it might have been tricky as I have issues submitted under `leastwood`

PR #324, opened 2021-10-11
New Colors 🎨

DianaPerkinsDesign:
- I left the orange hover color the same since there's no equivalent, and this one works with the new purple too. 
- There's a new button type: secondary CTA!


src/styles/hero.css
```diff
@@ -5,9 +5,12 @@
 }
 
 .hero h1 {
-  font-size: 80px;
-  text-shadow: 0 7px rgba(51, 0, 101, 1);
-  margin: 0 0 60px 0;
+  font-family: Roboto-Bold;
```
adamavenir, line 8: I think this needs to be `Roboto` and not `Roboto-Bold` (or else there's some other issue).

Changing to just `Roboto` fixes it though:

Screen Shot 2021-10-12 at 10 58 42 AM

src/styles/message-bar.css
```diff
@@ -1,30 +1,67 @@
...
   font-family: "Roboto", sans-serif;
-  border-top: 4px solid rgba(141,49,239,1);
-  border-bottom: 3px solid rgba(141,49,239,1);
-  opacity: .9
+  box-shadow: 0px 3px 20px 0px rgb(0 0 0 / 25%);
+  box-sizing: border-box;
 }
 
 .message-bar::before {
+  /* I broke this piece off so we could apply css filters/opacity to the background image if we wanted to */
```
adamavenir, line 14: nice

src/styles/base.css
```diff
@@ -1,43 +1,43 @@
...
+
+  --border-lightest-background: #71678a; /* border color. use with lightest background purple */
+
+  --text-headers: #fff;
+  --text-white: #fff;
+  --text-normal: #ffffffcc; /* cc = 80% */
+
+  /* these rgb color values should be identical to their pairs above */
+  /* use these to change opacity like this: `rgba(var(--main-purple-rgb), 0.8)` */
+  --main-purple-rgb: 132, 36, 226;
```
adamavenir, line 19: you're great @DianaPerkinsDesign. thanks for the useful comments and the nice consolidation cleanup here.

adamavenir (changes requested):
Found one issue with the hero typography that needs to be fixed. Looks awesome otherwise.

PR #489, opened 2021-11-17
Fetch readme HTML and social image from GitHub API

dstaley:
This PR makes a few small adjustments:

1. Instead of fetching the markdown of the contest repo's README, we instead fetch the GitHub generated HTML. This saves us from performing the markdown -> html conversion on the client-side. Plus it gets us neat things from GitHub's markdown parser, and ensures that what renders on the site matches what's on GitHub.
1. Instead of storing the contest graphics in the repo, we fetch them from GitHub. This makes it easier to add a contest since you don't need to put a specially named image in a specific location. Plus updating it is easier since you only need to upload a new image to GitHub.
1. Small style tweaks (removing `var/let` and using `const` instead, replacing empty render nodes with `null` instead of an empty string, etc.)
1. Removing some `console.log` statements from before the rebrand PR (figured now would be an okay time to remove them!)


DianaPerkinsDesign:
Thank you so much! It feels hilarious to me that there was generated html this whole time 😅

Issue #364, opened 2021-10-21
Use `<Link>` component for the links to contest reports


dstaley:
Currently we use a plain `<a>` element to link to the report page for a specific contest, which triggers a full page reload. We should switch to using `<Link>` so that we can benefit from the optimizations that Gatsby does automatically.

DianaPerkinsDesign:
Fixed with pr #523

PR #523, opened 2021-11-20
Replace links to internal pages with Gatsby Link

dstaley:
This PR replaces links to internal pages with Gatsby's `<Link>` component, which allows for client-side navigation instead of triggering a full roundtrip request to the server. This should make contest pages load even faster!


DianaPerkinsDesign (approved):
Closes #364 .  LGTM!

PR #634, opened 2021-12-09
Lots of little improvements

DianaPerkinsDesign:
Adds a link to the top section of a contest page if it has a published report.


DianaPerkinsDesign:
Converted to draft. I forgot to support `altUrl` when a report is a PDF or some other link.

src/layouts/ContestLayout.js
```diff
@@ -102,6 +96,26 @@ const ContestLayout = (props) => {
...
+              {props.data.markdownRemark &&
+              props.data.markdownRemark.frontmatter &&
+              props.data.markdownRemark.frontmatter.altUrl ? (
+                <Link
+                  to={props.data.markdownRemark.frontmatter.altUrl}
+                  className="button cta-button button-medium primary"
+                >
+                  View Report
+                </Link>
+              ) : null}
```
captainmangoC4, line 118: You can destructure `markdownRemark` similar to the approach with `contestsCsv`:
```
const { markdownRemark } = props.data;
```

It seems like markdownRemark could possibly be empty or undefined?
If so, you can define a variable to tell whether or not there is data for a report link (You can probably think of a better variable name for this):
```
const canViewReport = Boolean(markdownRemark.frontmatter)
```

```suggestion
              {canViewReport ? (
                <Link
                  to={markdownRemark.frontmatter.altUrl || `/reports/${markdownRemark.frontmatter.slug}`
                  className="button cta-button button-medium primary"
                >
                  View Report
                </Link>
              ) : null}
```

captainmangoC4 (approved):

DianaPerkinsDesign:
Add repo links to contest tiles also

src/components/reporter/Form.js
```diff
@@ -241,7 +238,7 @@ const Form = ({ contest, sponsor, repoUrl }) => {
               </div>
             )}
             {status === FormStatus.Submitted && (
-              <div>
+              <div style={{ textAlign: "center" }}>
```
captainmangoC4, line 241: Not a huge deal, but it's better to add a class here instead of inline styles

src/layouts/ContestLayout.js
```diff
@@ -94,14 +98,24 @@ const ContestLayout = (props) => {
...
                 >
                   Submit Finding
                 </Link>
               ) : null}
+              {canViewReport ? (
+                <Link
+                  to={reportUrl}
+                  className="button cta-button button-medium secondary"
+                >
+                  View Report
```
captainmangoC4, line 116: Did you know that instead of:
``` jsx
{condition ? <component/> : null}
```
you can do:
``` jsx
{condition && <component/>}
```
Just FYI. No need to change here. We use the former syntax in plenty of places. Plus, you could argue that the former is more readable. Maybe we can decide as a team which syntax we prefer.

src/layouts/ContestLayout.js
```diff
@@ -154,8 +168,17 @@ const ContestLayout = (props) => {
 };
 export default ContestLayout;
 
-export const pageQuery = graphql`
-  query ContestsId($contestId: Int) {
+export const myQuery = graphql`
```
captainmangoC4, line 171: Should we be naming these queries something more descriptive and connected to the component? Like maybe `contestLayoutQuery`? I'm not sure about good naming conventions for graphql.

src/styles/base.css
```diff
@@ -210,9 +210,7 @@ strong {
 }
 
 a {
-  text-decoration: none;
   color: #fff;
```
captainmangoC4, line 213: Should we have something else to identify links when they're not hovered? I may have misunderstood, but that's what I thought the original suggestion was about. The person was arguing that it's hard to find links on the page because they look like the rest of the text and you can only see them when you hover and they change color. Maybe we can keep our header and footer links like this, but make all other links (like in the body of reports) be a slightly different shade of orange when not hovered.

captainmangoC4 (approved):

src/layouts/ContestLayout.js
DianaPerkinsDesign: I DID NOT, that's super cool!

src/styles/base.css
DianaPerkinsDesign: Yup! I was removing `text-decoration: none`, so now they will all be underlined!

captainmangoC4: Disregard that comment, I didn't see at first that links are underlined. 😄

Issue #527, opened 2021-11-22
Add preview of markdown on submission form


DianaPerkinsDesign:
Let the wardens proof what they're writing before they send it. Suggested by warden Gerard Persoon.

PR #759, opened 2022-01-04
Add markdown preview to findings form

captainmangoC4:
#527


src/components/reporter/widgets/Widgets.module.scss
```diff
@@ -43,6 +43,30 @@
 }
 
 .Textarea {
-  height: 150px;
+  height: 250px;
   font-family: inherit;
+  resize: none;
```
DianaPerkinsDesign, line 48: We'll probably want to keep the resize since some wardens use this window to compose the message in, and we want to give them as much room as they need. It does break the layout, but in a way I accept as necessary for now 😄

DianaPerkinsDesign (commented):
Looking great; the wardens are going to be so happy to have this!

I left a small comment in the CSS, and I think two other things to possibly consider are: 
- Changing the font size to 13px (just for the input tab) to match the other inputs. This isn't a change you made, it's just something that we could fix while we're here 😄
- Adding some extra padding to the inside of the markdown preview tab, to let the text breathe a bit

src/components/reporter/widgets/Widgets.module.scss
captainmangoC4: Ok!

DianaPerkinsDesign (approved):

PR #774, opened 2022-01-05
Fixing time remaining = 0 bug 

DianaPerkinsDesign:
I'm hoping this will possibly solve needing to redeploy every contest start, but at the very least, it's a bug that's fixed now. 

`time.js`, `countdown.js`, and `contestTile.js` need to be rethought eventually. This is a patch to make this less painful in the present while we work on making those files read/work in a way we're happy with.


src/components/ContestTile.js
```diff
@@ -19,6 +19,8 @@ const ContestTile = ({ contest: { node } }) => {
   const { submissionPath, contestPath } = fields;
 
   const t = getDates(start_time, end_time);
+  let now = new Date().getTime();
+  const end = new Date(end_time).getTime();
```
captainmangoC4, line 23:
``` javascript
const hasContestEnded = () => {
   const now = Date.now();
   const contestEnd = new Date(end_time).getTime();
   return contestEnd - now >= 0;
}
```

src/components/ContestTile.js
```diff
@@ -66,7 +68,7 @@ const ContestTile = ({ contest: { node } }) => {
         >
           {`${findingsRepo === "" ? "Preview" : "View"} Contest`}
         </Link>
-        {t.state === "active" && findingsRepo && submissionPath ? (
+        {t.state === "active" && findingsRepo && submissionPath && (end - now >= 0) ? (
```
captainmangoC4, line 71:
``` javascript
{t.state === "active" && findingsRepo && submissionPath && hasContestEnded() ? 
```

captainmangoC4 (changes requested):

src/components/ContestTile.js
captainmangoC4: alternatively, you could omit this and just use `getTimeRemaining`

src/components/ContestTile.js
captainmangoC4: Now that I'm looking at the rest of the code, I realized you can import `getTimeRemaining` and then check that the total is > 0

captainmangoC4: So it would be
``` javascript
t.state === "active" && findingsRepo && submissionPath && getTimeRemaining().total > 0 ? 
```

captainmangoC4 (approved):

PR #812, opened 2022-01-10
Temporary work around for an issue with how component data is hydrated in Gatsby

### LINKED ISSUE #605: Deploys of the site sometimes result in bad cache data
Sometimes after an automatic deploy of the site, contest urls and images are mismatched. This is a serious problem because it can result in active contest pages not being accessible via links from the home page.

The easiest and thus far 100% effective method of resolving this issue is visiting [the site's Netlify build page](https://app.netlify.com/sites/code423n4-site/deploys?filter=main) and triggering a new deploy using the "Clear cache and deploy site" option.

Screen Shot 2021-12-04 at 7 45 44 AM

If it was possible to just force a cache clear on every netlify deploy (it doesn't seem to be?), that would absolutely resolve it -- but that's not ideal anyway as it eliminates the value of cacheing in the first place.

Based on my limited understanding, there seem to be multiple potential culprits:

- Netlify site — It may work to just completely delete the current setup and re-initialize a new netlify site config. this needs a bit of coordination though as it will take the site offline for a bit in order to remove the domain config associated with the old site config and set it up for the new one
- Gatsby cache — we might try updating to a newer version of gatsby or digging deeper into any known gatsby cache issues
- serviceworker config — browsers cache service workers for 24h and don’t update on reload. A serviceworker could be interfering somehow
- npm vs yarn — we are using yarn but netlify uses npm by default. I have seen some people talk about inconsistent results from netlify builds based on yarn. haven't reviewed whether netlify is actually running yarn -- it's supposed to if it detects a yarn.lock file.
- Netlify cache bug — maybe there's a bug here somewhere? (seems least likely)

Some references:

- [Netlify's options for custom cache headers](https://docs.netlify.com/routing/headers/)
- [HTTP cache control documentation](https://docs.w3cub.com/http/headers/cache-control)
- Harry's piece on [Cache Control for Civilians](https://csswizardry.com/2019/03/cache-control-for-civilians/) is a pretty thorough reference and explainer of these configuration options.

captainmangoC4:
fixes #605

### DIFF
--- a/src/components/ContestTile.js
+++ b/src/components/ContestTile.js
@@ -20,6 +20,8 @@ const ContestTile = ({ contest: { node } }) => {
   
   // @todo: Find a better solution to the below issue
   // This is a work-around for an issue with hydrating React components in Gatsby
+  // The drawbacks of this approach: the page momentarily renders contests that have
+  // expired and each contest tile component has to be rendered twice
   // This approach was suggested here: https://github.com/gatsbyjs/gatsby/discussions/17914
   const [isClient, setIsClient] = useState(false);
   
--- a/src/components/ContestTile.js
+++ b/src/components/ContestTile.js
@@ -1,7 +1,8 @@
-import React, { useState, useEffect } from "react";
+import React from "react";
 import { Link } from "gatsby";
 import Countdown from "./Countdown";
 import { getDates } from "../utils/time";
+import SponsorLink from './SponsorLink';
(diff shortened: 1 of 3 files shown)


DianaPerkinsDesign:
Very clever solution! Works well locally.

DianaPerkinsDesign (approved):
LGTM!

PR #827, opened 2022-01-11
Fix markdown preview bugs

### LINKED ISSUE #809: Issue submission preview tab: wrong preview of new lines
When submitting a finding, in the vulnerability details text field,
I am pretty sure that when entering just 1 new line between 2 lines, like this:
```
First line
Second line
```
The issue on Github will end up looking indeed like this:
```
First line
Second line
```

However, on the new preview tab (thanks 🙂), the above text would look like this:
```
First line Second line
```
---
So please verify that I'm correct -
that when writing vulnerability description, only 1 newline is needed between lines,
and if so,
the preview tab needs to be fixed to reflect this.

### LINKED ISSUE #810: Issue submission preview tab: fix overflowing code block
Hi,
When adding a long code block, the block's width is not being fitted to the form, but rather it overflows and messes the page design.

To reproduce (at least on my screen) add the following to the issue description:

    ```
    yesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyesyes
    ```
---
Note: while you're working on the CSS - the "create issue" button is missing `cursor: pointer` 😇

captainmangoC4:
resolves #810 
resolves #809

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,8 @@
     "react-select": "^4.3.1",
     "react-table": "^7.6.3",
     "react-tabs": "^3.2.3",
+    "remark-breaks": "^3.0.2",
+    "remark-gfm": "^3.0.1",
     "sass": "^1.32.12",
     "sharp": "0.28.2"
   },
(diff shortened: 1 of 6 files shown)


DianaPerkinsDesign (approved):
LGTM!

PR #828, opened 2022-01-11
Implement character limit for username in warden registration form

### LINKED ISSUE #796: Set character limit on warden registration form
Set a character limit for `handle` on the warden registration form. I don't currently have a suggestion for what the limit should be, but I would say to view all handles and see what the natural maximum seems to be. Perhaps `10`?

While we're in here, let's change the label from `handle` to `username`.

captainmangoC4:
resolves #796


DianaPerkinsDesign (approved):
Nice! Lots of good naming fixes in here.

PR #838, opened 2022-01-13
Add form validation to findings submission form

### LINKED ISSUE #763: Add client-side form validation to findings form
We should check that required fields are filled out before calling the function to submit findings. Without this, a warden can fill out all but one required field, submit the form, and then an error is thrown, which shows a vague error message that requires them to reload the page and lose all of the content that they filled out.

Side note -  the descriptive error message isn't shown, but it should be. When info is missing, instead of showing the error message from the function "Email, handle, address, risk, title, body, and labels are required." we just show "An error occurred"

captainmangoC4:
resolves #763


DianaPerkinsDesign (approved):
LGTM!

PR #868, opened 2022-01-18
Render sponsor link conditionally

### LINKED ISSUE #635: Support sponsor link being null
Currently, if the sponsor link field is not filled out, the site will still create an empty anchor tag. 

As part of fixing this issue, we should also fill out any missing sponsor links if we think we have the data.

captainmangoC4:
resolves #635 

Also fixes random build warnings


DianaPerkinsDesign (approved):
LGTM! I don't have to do a git blame to know all those `class=` were me 😅

PR #781, opened 2022-01-06
.nvmrc use 16.13.1

joshuashort:

PR #885, opened 2022-01-21
GitHub actions to node 16

joshuashort:
Pretty minor; gets automation tooling on same node major version.

Very related to #781.


DianaPerkinsDesign (approved):
This will be interesting, I don't know what happens when you edit a test that needs to test itself in order to merge 😄

LGTM!

PR #907, opened 2022-01-25
Make sure risk key is case insensitive

### LINKED ISSUE #881: Leaderboard data discrepancy for gas optimization counts
Warden `Pedroais` pointed out a discrepancy on the Leaderboard -- see this post: https://discord.com/channels/810916927919620096/859583659678105630/933787892864663582

I've checked and the findings.csv lists 4 gas opts for that warden, but the leaderboard says 0. 

It looks to me like some of them are correct (e.g. warden `nedodn`) but others are off, e.g. `jah`.

I have not investigated to see if the counts are off for other data displayed (winnings, all findings, high, med).

captainmangoC4:
In November 2021, we started using capital "G" for labeling risk as gas-optimization instead of lower case "g." The leaderboard only checked for the lower case label when counting findings.

Closes #881


DianaPerkinsDesign (approved):
Doing a double and reviewing this but also using your PR to confirm that my netlify bash stuff works properly 😂. 

LGTM!

PR #1088, opened 2022-02-18
GraphQL schema tweaks

joshuashort:
Some things that seem possible currently aren't in the GraphQL model..

Ok
* handles -> findings

Not ok
* [x] orgs -> contests
* [x] contests -> findings
* [x] findings -> contests


gatsby-config.js
```diff
@@ -99,6 +99,7 @@ module.exports = {
           start_time: csvDatetimeParser,
           end_time: csvDatetimeParser,
           hide: csvBooleanParser,
+          contest: `number`,
```
captainmangoC4, line 102: Would this replace contestid? Or do we need both for some reason?

joshuashort: No. Yes. It's a little tragic.

`contestid` is the name of the field in `contests.csv`
`contest` is the name of the field in `findings.csv`

This was the apparent way to do parsing on a column. We're just kinda lucky that we don't have enough columns out there to collide.. yet?

We could 'simply' change the column name in `findings.csv` to be `columnid`, and this bit'd be unnecessary.

captainmangoC4: Ah, I see. Ok, makes sense.

captainmangoC4 (approved):

PR #1091, opened 2022-02-18
Modify name code423n4 -> code4rena?

joshuashort:
There are other places -- when is it safe/desirable to do?


SamusElderg:
Closing stale PR

PR #1109, opened 2022-02-21
Create markdown files for QA and gas reports

captainmangoC4:
Whenever a QA or Gas report is submitted, instead of creating an issue with the contents of the report:
- create an issue with a link to the corresponding markdown file in the `data` directory
- add said markdown file with contents of the report to `data` directory
- include commit message that links back to the issue


functions/submit-finding.js
```diff
@@ -115,13 +121,17 @@ exports.handler = async (event) => {
   };
 
   try {
+    const markdownPath = `data/${handle}-${risk}-${Date.now()}.md`;
```
joshuashort, line 124: Maybe move this definition under `const path`, and:

```const markdownPath = `data/${handle}-${issueId}.md`;```

captainmangoC4: Good point. I actually planned on this first, but I wanted the issue to be able to link to the markdown file, and if I put the issue id in the markdown filename, then I can't know the path to the file until after the issue is created.

joshuashort: Oof. Yeah. Makes sense. Maybe there's a workaround.. because this makes programmatically getting the linked text a bit.. rough.

* Post the link as a comment?
* Alter the issue body with the link once the ID's known?

PR #1313, opened 2022-03-29
Make contest pages client only

captainmangoC4:


joshuashort (approved):
Any other places need this?

PR #1332, opened 2022-04-01
Fix error messages; allow spaces in discord usernames

captainmangoC4:


functions/request-support.ts
```diff
@@ -205,17 +191,16 @@ async function handler(event) {
       };
     }
 
-    const response = await notion.pages.create(body);
+    await notion.pages.create(body);
```
joshuashort, line 194: Is the response helpful? Are there cases where create can fail/not succeed, but not raise an exception?

joshuashort (approved):
Other question is optional -- looks good, since the Discord handle here doesn't mean the same as the warden handle, and doesn't need the same validation.

functions/request-support.ts
captainmangoC4: Not that I know of. I think I originally declared this because I thought I would return something from the response, but it wasn't used and it's not really necessary.

PR #1618, opened 2022-04-29
Contest artwork on go-live

joshuashort:


joshuashort:
Closing this in favor of #1666 -- also, the problem ended up being external repos not triggering a site rebuild, so it's deeper.

PR #1642, opened 2022-04-30
Add warden Leycun

code423n4:
This auto-generated PR registers the new warden Leycun


joshuashort:
No Discord announce (or maybe a different registration)

PR #1675, opened 2022-05-03
Add warden supernovahs

code423n4:
This auto-generated PR registers the new warden supernovahs


joshuashort:
Registered as #1676

PR #1851, opened 2022-05-17
Gatsby css warning

Simon-Busch:
-> Add webpack plugin to remove the css warning for order of import


joshuashort:
Let's hold on this 'fix' for now. The import ordering warning can be resolved, not silenced.

joshuashort:
> @joshuashort This is all good now :) I managed to find out where it came from and so deleted the webpack plugin !

Awesome -- thanks for digging in!

joshuashort (approved):
Much better without the warnings now!

PR #1985, opened 2022-05-25
Fix option background and font color

Simon-Busch:
- [x] Fix background color on option for leaderboard
- [x] Fix team avatar image  on leaderboard
- [x] Add clickable link on team


jack-the-pug:
It's great to see that team avatars will be visible on the leaderboard! Could you also add the link for teams?

Simon-Busch:
@jack-the-pug yep it will be better indeed ! 

Regarding the link for the teams, I need to find a way but it seems had, we decided to collapse the members of the team therefore, the whole div is supposed to be clicked to expand the team members.

I'll check if I can do something intuitive here

jack-the-pug:
@Simon-Busch 👍 Nice work! It looks really good now.

captainmangoC4 (approved):

PR #2002, opened 2022-05-25
Contest status

Simon-Busch:
- [x] Create new lambda netlify function
- [x] Create a function to fetch contestId && status from notionDb
- [x] map these with the data on the frontend
- [x] group contests by status
- [x] clean up logic
- [x] fix component calls 
- [x] Improve data fetching from notion API -> error catching
- [x] Refactor && optimize code
- [ ] Update front-end with design from @tealtigrrr 
- [x] Update gatsby config to pre-fetch all contests // status


Simon-Busch:
> Let's get this shipped. It's useful information to wardens even in the draft design state.

I agree

src/pages/contests/index.js
```diff
@@ -1,57 +1,197 @@
...
+        return 0;
+      });
+    }
+    return statusObject;
+  };
 
-  const filteredContests = contestsByState({ contests });
+  useEffect(() => {
+    setIsLoading(true);
+    fetch("/.netlify/functions/getNotionData")
```
joshuashort, line 89: Something like the following fits more with the current naming convention (to the extent that's a thing):

```suggestion
    fetch("/.netlify/functions/get-contest")
```

src/pages/contests/index.js
```diff
@@ -1,57 +1,197 @@
...
+        ...element.node,
+        status: statusObject[0]?.status,
+      };
+      switch (statusAndIdObj[0]?.status) {
+        case "Pre-Contest":
+          statusObject.upcomingContests.push(data);
+          break;
+        case "Preview week":
+          statusObject.upcomingContests.push(data);
+          break;
```
joshuashort, line 45: Can use fall-through.

```suggestion
        case "Pre-Contest":
        case "Preview week":
          statusObject.upcomingContests.push(data);
          break;
```

src/pages/index.js
```diff
@@ -18,7 +18,12 @@ export default function SiteIndex({ data }) {
 
   const contests = data.contests.edges;
   const filteredContests = contestsByState({ contests });
-
+  // TODO change this to fetch from notion
```
joshuashort, line 21: Doesn't it? And things seem to go weird with contest/node structure.

src/components/ContestList.js
```diff
@@ -7,7 +7,7 @@ const ContestList = ({ contests, updateContestStatus }) => {
       {contests.map((contest) => (
         <ContestTile
           contest={contest}
-          key={contest.node.id}
+          key={contest.id}
```
joshuashort, line 10: It doesn't seem necessary to do some of these plumbing changes.

```suggestion
          key={contest.node.id}
```

src/components/ContestTile.js
```diff
@@ -19,9 +19,8 @@ const ContestTile = ({ contest: { node }, updateContestStatus }) => {
     findingsRepo,
     repo: contestRepo,
     fields,
-  } = node;
-  const { submissionPath, contestPath } = fields;
-
+    status,
+  } = contest;
```
joshuashort, line 23: If going with the modifications.

```suggestion
  } = contest.node;
```

src/components/ContestTile.js
```diff
@@ -55,27 +54,30 @@ const ContestTile = ({ contest: { node }, updateContestStatus }) => {
           />
         ) : (
           <p>
-            Contest ran {t.startDay}-{t.endDay}
+            Contest ran {t.startDay === "" ? "" : t.startDay}-
```
joshuashort, line 57: It might make more sense to be more explicit about the.. state semantics? not (soon || active) might be unintentionally broad. It seems like it used to mean done, so a start/end day was guaranteed.

src/components/ContestTile.js
```diff
@@ -55,27 +54,30 @@ const ContestTile = ({ contest: { node }, updateContestStatus }) => {
...
+            {t.endDay === "" ? "" : t.endDay}
           </p>
         )}
         <ClientOnly>
           <Link
-            to={contestPath}
+            to={fields?.contestPath || "/"}
```
joshuashort, line 63: Similar context/semantics comment here -- `fields.contestPath` is practically mandatory.. and should it link to the main page when it doesn't exist? Seems like there'd be a broader issue.

src/components/ContestTile.js
```diff
@@ -55,27 +54,30 @@ const ContestTile = ({ contest: { node }, updateContestStatus }) => {
...
             className="contest-repo button button-small cta-button primary"
           >
             {`${findingsRepo === "" ? "Preview" : "View"} Contest`}
           </Link>
           {t.contestStatus === "active" && contestRepo && (
-            <a
-              href={contestRepo}
+            <Link
```
joshuashort, line 69: The `Link` tag is for internal links -- https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/#reminder-use-link-only-for-internal-links

src/components/ContestTile.js
```diff
@@ -55,27 +54,30 @@ const ContestTile = ({ contest: { node }, updateContestStatus }) => {
...
-            </a>
+            </Link>
           )}
-          {t.contestStatus === "active" && findingsRepo && submissionPath ? (
+          {(t.contestStatus === "active" || status === "Active Contest") &&
+          findingsRepo &&
+          fields.submissionPath ? (
             <Link
-              to={submissionPath}
+              to={fields?.submissionPath || "/"}
```
joshuashort, line 80: another: when submissionPath is expected, it should exist.

functions/_config.js
```diff
@@ -4,7 +4,9 @@ module.exports = {
   token: process.env.GITHUB_TOKEN,
   apiKey: process.env.MAILGUN_KEY,
   domain: process.env.MAILGUN_DOMAIN,
+  notionToken: process.env.NOTION_CONTEST_INTEGRATION_TOKEN,
+  notionDb: process.env.NOTION_CONTEST_DATABASE_ID,
```
joshuashort, line 8: Trying to move away from this _config (#2137); a bit more explicit, maybe.

```suggestion
  notionContestDb: process.env.NOTION_CONTEST_DATABASE_ID,
```

src/pages/contests/index.js
Simon-Busch: Totally ! I'll change it

src/pages/contests/index.js
Simon-Busch: Oh interesting I didn't know we can do it !

src/components/ContestTile.js
Simon-Busch: Oh didn't know ! Thanks

src/pages/index.js
Simon-Busch: Yes right I didn't do it here, I'll do it now

Simon-Busch: Done :-)

src/components/ContestList.js
Simon-Busch: It's important change actually because the object passer in contestList is now different

src/components/ContestTile.js
Simon-Busch: Same comment as above :)

src/components/ContestTile.js
Simon-Busch: Absolutely, that was a temp fix, don't remember exactly why but I had some errors, now that after re-test it seems good ! :)

src/components/ContestTile.js
Simon-Busch: Right, It cause an error for some reason.

So I think I can keep it this way knowing that that's part of the team's checklist to have this info today !
Capture d’écran 2022-07-12 à 06 12 49

src/components/ContestTile.js
```diff
@@ -59,23 +58,25 @@ const ContestTile = ({ contest: { node }, updateContestStatus }) => {
           </p>
         )}
         <ClientOnly>
-          <Link
-            to={contestPath}
+          <a
+            to={fields?.contestPath || "/"}
```
joshuashort, line 62: When would contestPath be undefined? Also,

```suggestion
            href={fields?.contestPath || "/"}
```

burgertime-c4 (approved):
LGTM 🚢

Issue #1431, opened 2022-04-12
Warden Auth


captainmangoC4:
- [x] implement client-side auth with Moralis
- [x] style user dropdown
- [x] implement team registration
- [x] allow merging addresses in moralis
- [x] verify user when they submit a finding while logged in
- [x] verify teams when a member submits a finding @captainmangoC4 
- [x] change UI in submission form to allow wardens who are on teams to toggle between submitting as an individual or as their team @captainmangoC4 
- [x] add a [+ email address] button to create multiple email fields for team submissions @captainmangoC4 
- [x] #2068
- [x] verify users when they register @captainmangoC4 
- [x] save polygon address for a team upon first submission @captainmangoC4 
- [x] create confirmation modal  @Simon-Busch 
- [x] use confirmation modal for confirming team polygon address and for merging addresses in moralis @captainmangoC4 
- [X] validate discord username @Simon-Busch 
- [x] protect finding submission page @captainmangoC4 
- [x] restyle submission form @captainmangoC4 
- [ ] create staging environment @joshuashort @captainmangoC4 
- [x] clean up finding submission form logic @captainmangoC4 
- [x] set up production database @captainmangoC4 
- [ ] finalize messaging @captainmangoC4 @tealtigrrr 
- [x] https://github.com/code-423n4/code423n4.com/issues/2046 -- style message for user to login for registration page and submit finding page
- [x] #2091
- [x] #2096
- [x] filter options in warden dropdown, removing wardens who have previously submitted findings?
- [x] clean up addresses from previous submissions table

captainmangoC4:
For the pages that require auth (submit findings and register team), if the user is not logged in, show a message telling them to login. Message should be styled to look like the submit finding success message.

For submit findings, it should say something like:

## Please log in
You need to be a registered warden currently connected via wallet to see this page.

**Note to existing wardens:**
As of [date] wardens are required to authenticate with a wallet to submit findings.

You can read more about this change here[link].

[connect with MetaMask] [connect with WalletConnect]



For team registration page, it should say:

## Please log in
You need to be a registered warden, currently connected via wallet to register a team.

[connect with MetaMask] [connect with WalletConnect]

Simon-Busch:
> For the pages that require auth (submit findings and register team), if the user is not logged in, show a message telling them to login. Message should be styled to look like the submit finding success message.
> 
> For submit findings, it should say something like:
> 
> ## Please log in
[...]

PR #2013, opened 2022-05-26
Create modal

Simon-Busch:
- [x] Create modal component
- [x] Create styling
- [x] Add x-icon
- [x] Update global import css


Simon-Busch:
@tealtigrrr  ready for review on design :-)

tealtigrrr:
Looks perfect! Great work! 🙌

Simon-Busch:
I will just add something before merging : 

- [ ] Add possibility to pass button text as props
- [ ] Add possibility to pass a second button ( which will replace the cancel button ) https://github.com/code-423n4/code423n4.com/issues/1431#issuecomment-1139141128

src/styles/modal.css
```diff
@@ -0,0 +1,86 @@
...
+  right: 13px;
+  height: 20px;
+  width: 20px;
+  background-color: transparent;
+  border: none;
+  cursor: pointer;
+}
+
+.modal-main-title {
+}
```
captainmangoC4: You can remove this

src/components/Modal.js
```diff
@@ -0,0 +1,39 @@
...
+        </div>
+        <div className="modal-main-buttons">
+          <button
+            className="button button-modal button-close"
+            onClick={handleClose}
+          >
+            Cancel
+          </button>
+          <button
+            className="button button-modal button-action "
```
captainmangoC4, line 46: Can you use existing button classes for these? like `cta-button` and `primary` or `secondary`

src/components/Modal.js
```diff
@@ -0,0 +1,39 @@
...
+        </div>
+        <div className="modal-main-content">
+          <p>{body}</p>
+        </div>
+        <div className="modal-main-buttons">
+          <button
+            className="button button-modal button-close"
+            onClick={handleClose}
+          >
+            Cancel
```
captainmangoC4: I like your idea to make the button text for both buttons dynamic. Maybe the props can be optional, and you can fall back to "Cancel" and "Ok" if the props are not provided.

Simon-Busch:
Right so here is how to call the component: 
Open state need to be handled in parent calling the component 
 
```javascript
[...]
  const [isOpen, setIsOpen] = useState(false);
[...]
<Modal
    show={isOpen}
    handleClose={() => setIsOpen(!isOpen)}
    title="test modal"
    body="hey Amy what do you think ?"
    primaryButtonAction={() => console.log("button 1")}
    primaryButtonText={'Button 1'}
    secondaryButtonAction={() => console.log("button 2")}
    secondaryButtonText={'Button 2'}
  />
```

src/styles/modal.css
```diff
@@ -0,0 +1,86 @@
...
+  background-color: var(--main-purple);
+}
+
+.display-block {
+  display: block;
+}
+
+.display-none {
+  display: none;
+}
```
captainmangoC4, line 76: instead of this, I would just add the class "modal" and set the style to `display: block` and then have another class "hidden" that is applied when `!shown` and that over-rides the style with `display: none`, or vice-versa.

src/components/Modal.js
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+
+const Modal = ({ handleClose, show, title, body, onAction }) => {
+  return (
+    <div className={show ? "modal display-block" : "modal display-none"}>
+      <div className="modal-main" onClick={handleClose}>
+        <button className="button-div modal-top">
+          <img
+            src="/images/x-icon.svg"
```
captainmangoC4, line 18: Can we use the same x-icon that we use for the mobile nav menu?

src/components/Modal.js
```diff
@@ -0,0 +1,57 @@
...
+              Cancel
+            </button>
+          ) : (
+            <button
+              className="button button-modal button-action"
+              onClick={secondaryButtonAction}
+            >
+              {secondaryButtonText}
+            </button>
+          )}
```
captainmangoC4, line 44:
```suggestion
            <button
              className="button button-modal button-close"
              onClick={handleClose}
            >
              {secondaryButtonText || "Cancel"}
            </button>

```

src/components/Modal.js
```diff
@@ -0,0 +1,57 @@
...
+          <button
+            className="button button-modal button-action "
+            onClick={primaryButtonAction}
+          >
+            {primaryButtonText}
+          </button>
```
captainmangoC4, line 50:
```suggestion
          <button
            className="button button-modal button-action "
            onClick={primaryButtonAction}
          >
            {primaryButtonText || "Ok"}
          </button>
```

captainmangoC4 (approved):

src/components/Modal.js
Simon-Busch: Done below :)

src/components/Modal.js
Simon-Busch: Done below :)

src/components/Modal.js
Simon-Busch: the x-icon in the mobile nav menu is css created and not svg if I am not wrong :)

src/components/Modal.js
Simon-Busch: Done in last commit :)

PR #2021, opened 2022-05-27
Auth for teams

captainmangoC4:
Verify findings submitted by teams
Allow wardens to submit as their team or as an individual


functions/submit-finding.js
```diff
@@ -68,15 +70,17 @@ exports.handler = async (event) => {
 
   // ensure we have the data we need
   if (
-    !email ||
-    !handle ||
+    !emailAddresses ||
+    !emailAddresses.length > 0 ||
```
joshuashort, line 74:
```suggestion
    !(emailAddresses.length > 0) ||
```

functions/submit-finding.js
```diff
@@ -99,15 +102,31 @@ exports.handler = async (event) => {
...
+      username: user,
     });
     if (!confirmed) {
       return {
         statusCode: 401,
         body: "Unauthorized",
       };
     }
+    if (attributedTo !== user) {
+      const url = `${event.headers.origin}/.netlify/functions/getUser?id=${attributedTo}`;
```
joshuashort, line 114: Does this work?

```suggestion
      const url = `/.netlify/functions/getUser?id=${attributedTo}`;
```

functions/submit-finding.js
captainmangoC4: good catch!

functions/submit-finding.js
captainmangoC4: Yes, it does. And in fact, node-fetch only allows using absolute urls. I can look into another work-around if this is not acceptable.

PR #2047, opened 2022-05-30
Update messaging for protected pages

Simon-Busch:
⚠️ jut add date and link before merging


Simon-Busch:
@captainmangoC4 Let me know this is OK for you :)

src/components/Modal.js
```diff
@@ -23,27 +23,29 @@ const Modal = ({
...
+              className="button cta-button primary"
+              onClick={primaryButtonAction}
+            >
+              {primaryButtonText || "Ok"}
+            </button>
+          </div>
+        ) : (
+          ""
+        )}
       </div>
```
captainmangoC4, line 49: I think you can undo the changes to this file. We won't have any modals without action buttons.

src/components/Login/Login.tsx
```diff
@@ -10,7 +10,7 @@ import Dropdown from "../Dropdown";
 import * as styles from "./Login.module.scss";
 import * as dropdownStyles from "../Dropdown.module.scss";
 
-const Login = () => {
+const Login = ({ dropDownViewOff }) => {
```
captainmangoC4, line 13: I would give this a name that reflects what it does show, vs what it doesn't show and then set it to false by default. For example: 
```javascript
const Login = ({ displayAsButtons = false }) => {
```
If we don't give it a default value, or define an interface for props where this is optional, then we will have typescript errors in the consuming components if they don't pass this prop.

src/components/reporter/SubmitFindings.js
```diff
@@ -438,7 +440,18 @@ const SubmitFindings = ({ wardensList, sponsor, contest, repo }) => {
       )}
     </div>
   ) : (
-    <p>You must connect your wallet to submit findings</p>
+    <div className="login-container">
+      <div className="finding-error-container">
+        <h1>Please login</h1>
+        <p>
```
captainmangoC4, line 446: Let's add a todo comment here to remind us that we need to include the official date once we know when this will be released

src/styles/forms.css
```diff
@@ -211,3 +211,25 @@ label {
...
+  width: 50vw;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  background-color: var(--secondary-background-purple);
+  border-radius: 8px;
+  border:1px solid var(--header-background-purple);
+  padding:20px;
+  box-shadow: 0px 4px 10px rgba(0,0,0,0.5);
+}
```
captainmangoC4, line 235: Is it possible to just utilize the className and styles that are applied to the form? e.g. from the Form module, `Form`

captainmangoC4 (changes requested):

src/styles/forms.css
captainmangoC4: You can also apply the `centered-text` class name

src/components/reporter/SubmitFindings.js
captainmangoC4: I see your note in the PR comment, but I think it's a good idea to add a comment in the code since this PR will be closed before the full feature is released.

src/components/Modal.js
Simon-Busch: Done :-)

src/components/Login/Login.tsx
Simon-Busch: Totally make sense ! I adapted it as well ✅

src/components/reporter/SubmitFindings.js
Simon-Busch: Yes correct so we make sure we change it before warden auth merge, Done ✅

src/styles/forms.css
Simon-Busch: Totally, Doesn't make sense to have duplicate classes, I changed it :) ✅

src/components/Dropdown.module.scss
```diff
@@ -38,7 +38,6 @@
 .Button {
   @include button;
   background: var(--dark-purple);
-  width: 100%;
```
captainmangoC4, line 41: This is actually necessary for the buttons in the dropdown. Otherwise they look like this:
image
But you can add another class to the non-dropdown buttons where you can over-ride this style if you need to. Alternatively, you could wrap the buttons in a div or something.

captainmangoC4 (approved):

PR #2066, opened 2022-05-31
Use modal component for confirmations

captainmangoC4:
@captainmangoC4


functions/submit-finding.js
```diff
@@ -35,12 +37,44 @@ async function getContestEnd(contestId) {
...
+    [`_data/handles/${teamName}.json`]: updatedTeamData,
+  };
+  const body = `This auto-generated PR adds polygon address for team ${teamName}`;
+  const title = `Add address for team ${teamName}`;
+  await octokit.createPullRequest({
+    owner: "code-423n4",
+    repo: "code423n4.com",
+    title,
+    body,
+    head: `warden/${teamName}`,
```
joshuashort, line 60: Make sure to add this to the ignore script (and/or nest another layer to ignore)

functions/submit-finding.js
```diff
@@ -107,46 +147,69 @@ exports.handler = async (event) => {
...
+            error: "You cannot submit findings for a team you are not on",
+          }),
         };
       }
+      if (!team.address) {
+        // create a PR to update team JSON file with team address
+        try {
+          await updateTeamData(team, address);
+        } catch (error) {
+          // don't throw error if this PR fails - there will likely be duplicates
```
joshuashort, line 180: This.. worries me.. but good notes to keep around for later review :)

captainmangoC4: Fair, but there are no real consequences to this failing. It's just trying to save the team's address for future submissions. I figured we wouldn't want to block a submission for this, even if it was a different failure. But I could at least check that the failure is due to a branch with that name already existing, and throw an error for any other failures.

functions/submit-finding.js
captainmangoC4: Oh, good catch. I misread the ignore script and thought it just checked for `warden` but now I see it looks for `warden-`

PR #2079, opened 2022-06-01
Verify registrations

captainmangoC4:


src/components/RegistrationForm.tsx
```diff
@@ -59,15 +59,19 @@ export default function RegistrationForm({
...
   const [state, setState] = useState(initialState);
   const [isNewUser, setIsNewUser] = useState(true);
   const [hasValidationErrors, setHasValidationErrors] = useState(false);
   const [isValidDiscord, setIsValidDiscord] = useState(true);
+
+  // global variables
   const avatarInputRef = useRef<HTMLInputElement>();
-  const { logUserOut } = useUser();
-  const { authenticate } = useMoralis();
   const regex = new RegExp(/.*#[0-9]{4}/, "g");
```
joshuashort, line 74: A little more descriptive name, maybe? And that first part is a bit of a problem.. Both are kind of find for now, but `.*` might be too permissive.

```suggestion
  const discordHandleRegex = new RegExp(/.*#[0-9]{4}/, "g");
```

joshuashort (approved):

src/components/RegistrationForm.tsx
captainmangoC4: Good point about `.*` being a bit permissive, but the goal here is just to make sure people don't accidentally enter their server nickname instead of their discord username. We're not validating or sanitizing any of the other inputs in the client.

PR #2081, opened 2022-06-02
Text area limit

Simon-Busch:
- [x] add limit to text area
- [x] Add span to indicate characters
EDIT : 
Seems that the max character for text area on issue submitting is not the problem here

Is seems it comes from the submitting of the form for HELP component 
-> Limited **by notion API** to 2000 chars


captainmangoC4:
You might have already been planning on this, but just in case I'll suggest adding a note suggesting that users create a gist and paste the link in the textarea if they need to submit something larger than 2000 characters.

Simon-Busch:
Yes at the moment what I added is a maxLenght for help form and a span when it comes characters comes close 2000.

Capture d’écran 2022-06-03 à 06 33 27

Simon-Busch:
@tealtigrrr Would you see the span differently ?

src/components/reporter/widgets/TextArea.js
```diff
@@ -12,25 +12,46 @@ const TextArea = ({ name, required, fieldState, isInvalid, onChange }) => {
     onChange(e);
   }
 
+  const checkPageLocation = () => {
+    if (typeof window !== `undefined`) {
+      return window.location.href.includes("/help");
+    }
+  };
```
captainmangoC4, line 19: Instead of adding this specific logic into the TextArea component (the TextArea should not have to be aware of the location) let's add an optional prop to this component for `maxSize` and pass that from the `help` consumer.

captainmangoC4 (changes requested):

src/components/reporter/widgets/TextArea.js
Simon-Busch: I see, I'll adapt accordingly

captainmangoC4 (approved):

src/pages/help.js
```diff
@@ -49,6 +49,7 @@ function ContactUs() {
       name: "description",
       label: "Description",
       widget: "textarea",
```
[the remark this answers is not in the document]
Simon-Busch: Yep totally !! 

I just added it =)

src/pages/help.js
```diff
@@ -49,7 +49,9 @@ function ContactUs() {
...
+      maxSize: 2000,
       required: true,
+      helpText: "For submission longer than 2000 characters, add a link to a gist"
```
captainmangoC4, line 54:
```suggestion
      helpText: "For content longer than 2000 characters, please include a link to a gist in GitHub"
```

captainmangoC4 (approved):

PR #2093, opened 2022-06-03
Add toaster on login promises

Simon-Busch:
- [x] add toaster to handle promises


Simon-Busch:
@captainmangoC4 ready for your review, I was not sure about other places where this toast would be needed actually

src/components/Login/Login.tsx
```diff
@@ -37,7 +37,16 @@ const Login = ({ displayAsButtons = false }) => {
...
 
     try {
-      await connectWallet();
+      const connectWalletHandler = new Promise(resolve => {
+        resolve(connectWallet());
+      });
+
+      toast.promise(connectWalletHandler, {
+        pending: "Logging in",
+        success: "Logged in ✅",
```
captainmangoC4, line 46: I would remove these icons with the success and error messages because the toast notification should already have an icon in the top left corner.

captainmangoC4 (changes requested):

src/components/Login/Login.tsx
```diff
@@ -37,7 +37,16 @@ const Login = ({ displayAsButtons = false }) => {
...
+        error: "Something went wrong ❌",
```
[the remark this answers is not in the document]
Simon-Busch: You can look at the updated commit, I found that it's actually much easier and makes more sense to have an initial toaster that evolves with the different promises ( authenticate || connectWallet ) and attribute it the possible errors

functions/get-team.ts
```diff
@@ -33,7 +33,7 @@ exports.handler = async (event) => {
       };
     } else {
       return {
-        statusCode: 401,
+        statusCode: 200,
```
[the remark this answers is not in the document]
Simon-Busch: I actually updated it to a status 204.

I think it would makes sense as it would mean that the call was successfull but there is no body to the request. What do you think ?

src/components/Login/Login.tsx
```diff
@@ -19,49 +19,87 @@ const Login = ({ displayAsButtons = false }) => {
...
+        autoClose: 2000,
+      });
       return;
     }
 
     try {
-      await connectWallet();
+      const connection = await connectWallet();
+      console.log(connection);
+      if (connection) {
```
captainmangoC4, line 52: I think you can remove this. See comment on line 59-66.

src/components/Login/Login.tsx
```diff
@@ -19,49 +19,87 @@ const Login = ({ displayAsButtons = false }) => {
...
+          autoClose: 2000,
+        });
+      } else {
+        toast.update(id, {
+          render: "Please register as warden",
+          type: "error",
+          isLoading: false,
+          autoClose: 2000,
+        });
+      }
```
captainmangoC4, line 66: I think we can exclude this part. `connectWallet()` doesn't return anything. And it throws any errors that are encountered, so the `catch` block below will handle the different error states. It's a good idea to update the toast message when a user isn't registered though. To do this, I'd add another error type to `UserLoginError` enum and throw that from `connectWallet()` right before the user is redirected to the registration page. I'd give it the "info" type instead of "error" though. Although, I'm not sure if the toast messages persist when redirected. You'll have to test that.

src/components/Login/Login.tsx
```diff
@@ -19,49 +19,87 @@ const Login = ({ displayAsButtons = false }) => {
...
     event.preventDefault();
+    const id = toast.loading("Logging in...", { autoClose: 2000 });
+
     try {
       const user = await authenticate({ provider });
       if (user === undefined) {
         // user clicked "cancel" when prompted to sign message
         // @todo: update messaging
-        toast.error("You must sign the message to connect your wallet");
+        toast.update(id, {
```
captainmangoC4, line 29: Clever idea! This seems like it will work perfectly!

captainmangoC4 (changes requested):

src/components/Login/Login.tsx
captainmangoC4: Also, I'd go ahead and remove the console.log

src/components/Login/Login.tsx
Simon-Busch: Actually, I tried to remove it and so, only the "logged in" pops.

Actually to manage this I added a final return statement to connectWallet : "connected" if the whole process is sucessfull, that's how I get it with :

```javascript
      if (connection) {
        toast.update(id, {
          render: "Logged in",
          type: "success",
          isLoading: false,
          autoClose: 2000,
        });
      } else {
        toast.update(id, {
          render: "Please register as warden",
          type: "error",
          isLoading: false,
          autoClose: 2000,
        });
      }
      ```
      
      But I can definitely remove the this block : 
      ```javascript 
      else {
        toast.update(id, {
          render: "Please register as warden",
          type: "error",
          isLoading: false,
          autoClose: 2000,
        });
      }
      ```
      As it won't persist upon navigation anyway

captainmangoC4 (approved):

PR #2099, opened 2022-06-03
Re-work reset registration form

Simon-Busch:
- [x] update register component -> change to modal
- [x] update modal component and context
- [x] pass resetForm to registration form component


Simon-Busch:
@captainmangoC4 I took the liberty to adapt the component to use our great modal, I think it makes sense in the case, let me know what you think about it :-)

src/pages/register.tsx
```diff
@@ -45,42 +47,47 @@ export default function UserRegistration({ data }) {
...
+          <p>An error occurred while processing your registration.</p>
+          {errorMessage !== "" && (
+            <p>
+              <small>{errorMessage}</small>
+            </p>
+          )}
+        </>
+      ),
+    });
+  };
```
captainmangoC4, line 72: Apologies - it seems we had a miscommunication about this. We should actually show the error message on the page, like we do with error messages on other forms, not in a modal.

The issue with the form state resetting when an error occurs is due to the fact that the error state is managed outside of the form component. The `register` page component manages the error state, so when an error occurs, the `RegisterForm` component is unmounted and the error message is shown instead. Then, when the error state is reset and we show the form again, it is actually remounting the form component with blank state. If we either 1. manage error state inside `RegisterForm`, or 2. remove `RegisterForm` altogether and move its contents directly into `register` page component, we can solve this problem.

captainmangoC4 (changes requested):

captainmangoC4 (approved):

PR #2173, opened 2022-06-13
Fix style of contest artwork

Simon-Busch:
- [ ] update css style


hannahtuttle (approved):
@Simon-Busch I tested this and it looks good!

Issue #2187, opened 2022-06-14
Arbitrary file can be read


csanuragjain:
In get-user.ts, the passed user handle id is not sanitized. Attacker can simply use below URL to access internal directories

${event.headers.origin}/.netlify/functions/get-user?id=${handle} where malicious id can be like ../../some_other_file.json

## Problem:

const userFile = readFileSync(`./_data/handles/${userHandle}.json`); is directly trusting the value of userHandle where const userHandle = event.queryStringParameters.id; Since this value is not sanitized so attacker can easily give malicious path

## Recommendation
Use the isDangerousHandle(id) function to get rid of malicious input which is used currently in other functions

csanuragjain:
Just to add, the url need to be accessed using Proxy Interceptor like Burp Suite which will not process .. before sending it to server. As a shortcut even Mozilla Firefox edit and resend feature can be used in network console to send this request

eugenioclrc:
Ok so what if we put the functions `isDangerousHandle` and `isDangerousRepo` in a helper utilities file like `_utils.js` so we could share the function logic.
Then we could import it on [`users.ts`](https://github.com/eugenioclrc/code423n4.com/blob/14c4ba945235ffd81359c534220e2d6464f4165d/functions/users.ts#L2)

csanuragjain:
yes I think that is a good idea

Simon-Busch:
@captainmangoC4, is this issue still on ?  Or has it been done in another branch ? :-) 

I saw that the PR(https://github.com/code-423n4/code423n4.com/pull/2195)  from @eugenioclrc  was closed ?

captainmangoC4:
@Simon-Busch thanks for asking! I forgot what this issue was originally about. Yeah, it is still open. We need to add the same “isDangerous” check in get-user and get-teams.

PR #2195, opened 2022-06-14
Check userhandle refactor

eugenioclrc:
Fix to issue [2187](https://github.com/code-423n4/code423n4.com/issues/2187)

Plus a refactor to avoid duplicate isDangerous function.

### DIFF
--- a/functions/_utils.js
+++ b/functions/_utils.js
@@ -1,0 +1,3 @@
+export function isDangerous(s) {
+  return s.match(/^[0-9a-zA-Z_\-]+$/) === null;
+}
--- a/functions/_utils.js
+++ b/functions/_utils.js
@@ -1,3 +1,0 @@
-export function isDangerous(s) {
-  return s.match(/^[0-9a-zA-Z_\-]+$/) === null;
-}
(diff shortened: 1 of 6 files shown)


captainmangoC4 (changes requested):
Unfortunately netlify lambda functions don't make it this easy to share code across multiple functions. I've seen people in [netlify forums](https://answers.netlify.com/t/sharing-helper-function-between-lambdas/13934) suggesting to make utils into an npm package as a way around this. Another option is to call one lambda function from another with a HTTP request (you can see an example of this [here](https://github.com/code-423n4/code423n4.com/blob/feature/warden-auth/functions/submit-finding.js#L141))

Both of those options seem like overkill for a simple regex check. Maybe there is another option I haven't come across though.

eugenioclrc:
I have remove the shared js and import the function on each endpoint, with a comment explaing why the function is repeated.

captainmangoC4:
It looks like you made these changes to an out-of-date fork. The file `users.js` was replaced with `get-user.js`. I'm not sure what would be the best way to resolve the conflicts - I'm not a git expert. @eugenioclrc Is it alright with you if I just copy over the work into a fresh branch?

eugenioclrc:
This pull request is not valid anymore.

eugenioclrc:
Yikes! Agree, I will rollback or at least try to find a suitable solution

PR #2261, opened 2022-06-18
Add warden Vin007

code423n4:
Auto-generated PR to register the new warden Vin007

I wrote contract in solidity and have very knowledge in auditing.


SamusElderg:
User-submitted wrong name, have clarified and they will re-apply with the correct handle + more information about experience level, closing this PR

PR #2293, opened 2022-06-21
Create basic manage account form

captainmangoC4:


Simon-Busch (approved):
All good, We really need to refactor these forms later we have a few similar pages with the exact duplicates of code ... That was the idea of the unique form ^^

PR #2313, opened 2022-06-23
Add warden LeoGold

code423n4:
Auto-generated PR to register the new warden LeoGold

I have experience reading smart contract to understand and at the same time looking for codes that won't work according to the author's intention,  I was also able to ransacked the evm storage to break into a smart that was not verified in a cf challenge.
I help colleagues read their codes and prevent them from deploying codes that are won't execute according to their intention.
I love to read code a lot because it is what I enjoy


0xSorryNotSorry (approved):

0xSorryNotSorry:
Approved

PR #2330, opened 2022-06-24
Check if user Handle is malicious

Simon-Busch:
- [x] Add isDangerousHandle
- [x] Check userHandle and return error if it is
- [x] Update get-user && get-team

### DIFF
--- a/functions/get-team.ts
+++ b/functions/get-team.ts
@@ -22,7 +22,7 @@ exports.handler = async (event) => {
     return {
       statusCode: 400,
       body: JSON.stringify({
-        error: "Handle can only contain alphanumeric characters [a-zA-Z0-9], underscores (_), and hyphens (-).",
+        error: "Team name can only contain alphanumeric characters [a-zA-Z0-9], underscores (_), and hyphens (-).",
       }),
     };
   }
--- a/functions/get-team.ts
+++ b/functions/get-team.ts
@@ -11,7 +11,21 @@ exports.handler = async (event) => {
       headers: { Allow: "GET" },
     };
   }
+
+  function isDangerousHandle(s) {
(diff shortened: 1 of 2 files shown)


Simon-Busch:
@captainmangoC4 When you have time, can you review this and merge if fine ? :)

functions/get-team.ts
```diff
@@ -11,8 +11,22 @@ exports.handler = async (event) => {
       headers: { Allow: "GET" },
     };
   }
+
+  function isDangerousHandle(s) {
+    return s.match(/^[0-9a-zA-Z_\-]+$/) === null;
```
0x1f8b, line 16: Seems that in line https://github.com/code-423n4/code423n4.com/pull/2330/files#diff-f56b9b0742f3b1b01343dbf9754332291dcf6d5c81644f6fd6c5af840d9563deR35 it will require the ".json"

shargon: Yes, you are right

0x1f8b: Yes, you're right, I mixed up the names with the phone.

functions/get-team.ts
```diff
@@ -11,8 +11,22 @@ exports.handler = async (event) => {
...
+  }
+
   const userHandle = event.queryStringParameters.id;
 
+  if (isDangerousHandle(userHandle)) {
+    return {
+      statusCode: 400,
+      body: JSON.stringify({
+        error: "Handle can only contain alphanumeric characters [a-zA-Z0-9], underscores (_), and hyphens (-).",
```
captainmangoC4, line 25:
```suggestion
        error: "Team name can only contain alphanumeric characters [a-zA-Z0-9], underscores (_), and hyphens (-).",
```

captainmangoC4 (approved):

PR #2895, opened 2022-07-13
2762 feature self serve issue editing populate form

captainmangoC4:


src/components/TeamRegistrationForm.tsx
```diff
@@ -44,30 +45,25 @@ function getFileAsBase64(file) {
...
   const { currentUser } = useUser();
   const { user, isInitialized } = useMoralis();
 
   const [state, setState] = useState(initialState);
-  const [hasValidationErrors, setHasValidationErrors] = useState(false);
-  const [status, setStatus] = useState<FormStatus>(FormStatus.Unsubmitted);
-  const [errorMessage, setErrorMessage] = useState<string | ReactNode>("");
   const [teamMembers, setTeamMembers] = useState<
-    { value: string; image: string }[]
+    { value: string; image: unknown }[]
```
joshuashort, line 57: Something like this, for the string field that might not exist?

```suggestion
    { value: string; image: string | undefined }[]
```

src/components/reporter/findings/fields.js
```diff
@@ -1,14 +1,14 @@
 export const emailField = {
   name: "email",
-  label: "Email address",
+  label: "Email address *",
```
joshuashort, line 3: Is it possible to accomplish this a different way? CSS + required?

```suggestion
  label: "Email address",
```

src/pages/register-team.tsx
```diff
@@ -4,12 +4,12 @@ import React from "react";
...
-import * as styles from "../components/form/Form.module.scss";
-
 export default function TeamRegistration({ data }) {
-  const handles = new Set(data.handles.edges.map((h) => h.node.handle));
+  const handles: Set<string> = new Set(
+    data.handles.edges.map((h) => h.node.handle)
+  );
 
-  let wardens: { value: string; image: string }[] = [];
+  let wardens: { value: string; image: unknown }[] = [];
```
joshuashort, line 12: Similar question -- probably has other effects

```suggestion
  let wardens: { value: string; image: string | undefined }[] = [];
```

joshuashort:
So many good changes 🙏

src/components/TeamRegistrationForm.tsx
captainmangoC4: The reason I changed it to `unknown` instead of `string` is actually because it's not a string. It's an object and the image src itself is `image.childImageSharp.resize.src`. I didn't feel like defining all of that, so I just made it `unknown`.

joshuashort: Oops, it is actually an image then -- ok! :)

src/components/reporter/findings/fields.js
captainmangoC4: Good idea! I added it to the `Input` component and left a comment in this file to delete them once we migrate to using the `Input` everywhere.

src/pages/register-team.tsx
captainmangoC4: Yeah, same here. We could define the Sharp Image object, but I thought it was fine as an intermediate step (where we have some typescript sprinkled in with javascript) to just say the type is unknown. What do you think?

joshuashort (approved):

src/pages/register-team.tsx
joshuashort: Later's fine -- I misunderstood what these were.

PR #2992, opened 2022-07-19
Create list of findings that link to report form

captainmangoC4:


joshuashort (approved):
So good -- nice approach(es) and todo/notes!

PR #3023, opened 2022-07-22
2762 feature self serve issue editing frontend

captainmangoC4:


src/components/reporter/SubmitFindings.tsx
```diff
@@ -2,6 +2,10 @@ import clsx from "clsx";
 import React, { useEffect, useState, useCallback } from "react";
 import { toast } from "react-toastify";
 
+// types
+import { Field } from "./widgets/Widgets";
+import { FindingRequestData, ReportState } from "../../templates/ReportForm";
```
joshuashort, line 7: This import seems weird -- probably that there's something that might not belong in that templates location?

joshuashort: .. yeah, it seems these things function like types.

If that makes sense, they can probably be moved into `types`, but doesn't have to be for this PR.

joshuashort:
Very good set of improvements / fixes!

PR #3374, opened 2022-08-09
Register warden solo

code423n4:
Auto-generated PR to register the new warden solo


SamusElderg:
No member in the C4 Discord channel with a matching handle

PR #3540, opened 2022-08-19
2764 feature custom payment address

captainmangoC4:


captainmangoC4:
This should be merged at the same time as https://github.com/code-423n4/buttons/pull/233

PR #3674, opened 2022-08-30
added warning on Judge Application

hannahtuttle:
…


src/pages/judge-application.js
```diff
@@ -205,6 +209,10 @@ const JudgeApplication = () => {
                         </p>
                       </div>
                     )}
+                    {isLoggedInErrorMessage ? <p 
+                    className = "error-message"
+                    > User must be logged in to submit this form. 
+                    </p> : ''}
```
captainmangoC4, line 215: Since this page should only be interacted with by a logged-in user, we can just wrap it in a `ProtectedPage` component instead of `DefaultLayout`.  (see certified-contributor-application as an example)

captainmangoC4 (changes requested):

hannahtuttle:
@captainmangoC4 I updated this to have a protected page instead of a default layout. Can you check this again?

captainmangoC4 (approved):

PR #3854, opened 2022-09-10
Register warden volky

code423n4:
Auto-generated PR to register the new warden volky

@volkyeth


SamusElderg:
User exists but has not yet checked into the i-want-to-be-a-warden channel

SamusElderg (approved):

PR #3993, opened 2022-09-19
Change handle to 100proof

sseefried:
I gone through and changed my handle in all the appropriate places. I've even gone through the reports and replaced the links to my private forks of repositories. They now point and new repos in another GitHub (one-hundred-proof).


kartoonjoy:
As discussed with the warden, closing this PR.

PR #4012, opened 2022-09-20
Nighthawk/231 pitch issue categorization

hannahtuttle:
The labels that are added onto each repo needs to be updated before this is ready so that when when the classification labels are selected during a contest the labels update correctly.


src/templates/ReportForm.tsx
```diff
@@ -66,7 +68,7 @@ const ReportForm = ({ data, location }) => {
   // state
   const [state, setState] = useState<ReportState>(initialState);
   const [attributedTo, setAttributedTo] = useState<string>("");
-  const [isLoading, setIsLoading] = useState<boolean>(true);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
```
captainmangoC4, line 71: This was actually set to true by default for a purpose. We want to show a loading state until the finding details are fetched when we're editing findings. This same page is used for submitting and editing findings. If it's set to false by default, then we'll see the form populate with data for a new finding and then abruptly update when the finding details are fetched.

captainmangoC4:
Let's prefix these labels with something like "classification-" so that they are easily parseable for de-duping purposes.

src/templates/ReportForm.tsx
hannahtuttle: Good catch, I changed that for testing something, I removed all my testing code, but I missed this! I will change it back now!

SamusElderg:
@hannahtuttle & @captainmangoC4 can we close this one off? doing a bit of a cleanup of PRs in this repo

SamusElderg:
Closing stale PR

PR #4066, opened 2022-09-22
Register warden korey-corkscrew

code423n4:
Auto-generated PR to register the new warden korey-corkscrew

@korey-corkscrew


SamusElderg:
The user exists but has not yet checked into the `i-want-to-be-a-warden` channel, closing PR for now

PR #4127, opened 2022-09-26
Show message on contest page for "versus" contests

captainmangoC4:
@sockdrawermoney @CloudEllie can you both review the messaging for this?
Here's a snapshot of how it looks on the website:
image


sockdrawermoney:
Actually, I think in terms of information leakage, it would be ideal if it 404'd :) Sorry for the extra work.

I learned this from Adam Baldwin when ^lift audited GitHub a million years ago. You can thank him for the fact that GitHub returns 404s instead of any info about repos you don't have permission to access. (If it was an access rights issue and not a 404, GitHub used to say something that implied you needed to check that you had proper permissions, but he pointed out that could be used to leak sensitive info about projects.)

captainmangoC4:
No worries, it wasn't much work at all. So you think it should just say "404 not found" on the contest page?

captainmangoC4:
Or you're saying we should not create the page at all if it's a private contest (so the link would go to our 404 page)

captainmangoC4:
I think we can omit option 1 and change the title to just say something about how this is not an open contest.

joshuashort (approved):

joshuashort:
This one? Or #5163?

burgertime-c4 (approved):

PR #4179, opened 2022-09-29
Custom payment addresses, and login with username and password

captainmangoC4:


joshuashort (approved):

Simon-Busch:
Just a quick question regarding the custom payment addreses, I tested the PR locally and these fields are empty, isn't the polygon address  supposed to be populated by default ??

Capture d’écran 2022-10-04 à 06 19 00

Also, for my understanding, if I am a new warden and want to login with a wallet the form is exactly the same for both option, it is normal ?

Capture d’écran 2022-10-04 à 06 21 10

Other that, seems great :)

Simon-Busch (approved):

captainmangoC4:
> Also, for my understanding, if I am a new warden and want to login with a wallet the form is exactly the same for both option, it is normal ?

I'm not sure what you mean. Are you referring to the fact that password is required for both? The main differences between the two tabs are payment address field and submit button(s).

`Register with wallet` tab has an option to use the auth address you sign up with as your polygon payment address. If you uncheck the box, then you'll be required to enter a payment address. It also has 2 buttons for the metamask and walletconnect.

`Register with password` tab has payment address as a required field because there is no auth address. Also, there is 1 register button.

There is an option for users to set/reset their passwords after registering with a wallet, so we could make the password field optional for registration with wallet if that make more sense. I just felt like this made it obvious to the user when they're registering that they will have the option to login with username and password. If we remove it, they might not discover that they have another option for login.

PR #4257, opened 2022-10-06
Narrow contest status bins based on contest start/end

captainmangoC4:


captainmangoC4:
Currently, we're only organizing contests based on upcoming/active/completed statuses, but I left in the switch statements for the future in case we decide to show more granular statuses. That's why there are switch statements that all fall through to the default.

Simon-Busch (approved):

PR #4315, opened 2022-10-09
Register warden darth_cy

code423n4:
Auto-generated PR to register the new warden darth_cy


SamusElderg:
The user exists but has not yet checked into the `i-want-to-be-a-warden` channel, leaving open for now

SamusElderg (approved):

PR #4346, opened 2022-10-11
Fix timing on contests

Simon-Busch:


Simon-Busch:
Resolved in https://github.com/code-423n4/code423n4.com/pull/4346

PR #4373, opened 2022-10-12
Fix broken layout for numbers in contests page

Simon-Busch:


joshuashort:
This just disables Math processing -- a feature that was added by request. We could use better error handling/guidance around report generation.

It might be better to change the allowed escape characters for the math processor.

joshuashort:
Preserving for now.

PR #4384, opened 2022-10-13
Register warden derked

code423n4:
Auto-generated PR to register the new warden derked

@Derked


Derked:
I was trying to re-register. I had never authenticated with the wallet.

SamusElderg:
> 

Oh perfect, yes that makes sense! Thanks!

SamusElderg (approved):

PR #4553, opened 2022-10-23
Register warden LeLe

code423n4:
Auto-generated PR to register the new warden LeLe

@790228724


SamusElderg:

SamusElderg:
Scratch that, looks like its this user: `Musk#1407` based on the PFP, have contacted them to change their nickname to match the application

SamusElderg:
User has not responded to name-change, closing PR

PR #4590, opened 2022-10-25
Register warden Jules

code423n4:
Auto-generated PR to register the new warden Jules


SamusElderg:

SamusElderg:
User will also need to change their nickname to `Jules` instead of `Jules |` if/after they check into the channel

SamusElderg (approved):

PR #4689, opened 2022-10-28
Register warden web3

code423n4:
Auto-generated PR to register the new warden web3

@web3proof


SamusElderg:
Cant find a matching user on Discord, however they have a PFP so might be able to match up that way after finalising other applications

PR #4699, opened 2022-10-28
Register warden Kamil93

code423n4:
Auto-generated PR to register the new warden Kamil93


SamusElderg:
I think might be this user: `kamil166#9011` have not contacted to confirm yet

PR #4710, opened 2022-10-28
Register warden Abdo

code423n4:
Auto-generated PR to register the new warden Abdo

@https://twitter.com/ELYOUSFI1992?t=IWSAHLgcTsJK9zJGuZQdcA&s=09


SamusElderg:
Could be `abdo_ssamad#6938` or `abdoabdo#6904` but i think we need to PM this user and see why they uploaded a passport photo, bit worried they made a mistake

PR #4746, opened 2022-10-28
Register warden LEVI

code423n4:
Auto-generated PR to register the new warden LEVI

@LEVI0087


SamusElderg:
Might be this user `Levi ($POX)#4505` but have not contacted them yet to confirm

PR #4799, opened 2022-10-28
Register warden aspro

code423n4:
Auto-generated PR to register the new warden aspro


SamusElderg:
Might be user: `! ASPRO ($POX)#7704` but have not contacted them to check yet

PR #4801, opened 2022-10-28
Register warden Hamza

code423n4:
Auto-generated PR to register the new warden Hamza

@ Ham9ZA


SamusElderg:
Might be user: `Hamzamostafi#4274` or `Hamzacrypto99#0554` will need to clarify after other applications are completed

PR #4803, opened 2022-10-28
Register warden fulgence1O

code423n4:
Auto-generated PR to register the new warden fulgence1O

@fulgence10


SamusElderg:
Must be user: `fulgence10#9711` *NOTE* that is a capital 'o' not a number '0' in their application so will need to update their nickname after they check into the channel

PR #4832, opened 2022-10-29
Register warden ahmedshanto1999

code423n4:
Auto-generated PR to register the new warden ahmedshanto1999


SamusElderg:
The user exists but has not yet checked into the `i-want-to-be-a-warden` channel || Probably: `Ahmed shanto#2731` will have to change nickname to match if so

PR #4839, opened 2022-10-29
Register warden kaka

code423n4:
Auto-generated PR to register the new warden kaka


kkmmmo:
谢谢

SamusElderg:
Maybe user: `kaka8989#5784` but has not checked into the `i-want-to-be-a-warden` channel so will confirm after the pending applications are completed

PR #4863, opened 2022-10-29
Register warden uttam665

code423n4:
Auto-generated PR to register the new warden uttam665


SamusElderg:
The user is probably: `uttam177#5743` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #4935, opened 2022-10-29
Register warden Rahmat

code423n4:
Auto-generated PR to register the new warden Rahmat


SamusElderg:
The user is probably: `rahmat.rm#8926` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #4942, opened 2022-10-29
Register warden SHUAI

code423n4:
Auto-generated PR to register the new warden SHUAI


SamusElderg:
The user is probably `SHUAI |cega#9271` but has not yet checked into the `i-want-to-be-a-warden` channel anyway

PR #4953, opened 2022-10-29
Register warden Rudeus

code423n4:
Auto-generated PR to register the new warden Rudeus


SamusElderg:
User is probably `Rudeusiu#9767` but have not contacted them to find out yet

PR #4978, opened 2022-10-29
Register warden farmin

code423n4:
Auto-generated PR to register the new warden farmin


SamusElderg:
The user is probably `farmin021#6783` but has not yet checked into the `i-want-to-be-a-warden` channel anyway

PR #4994, opened 2022-10-29
Register warden smit_rajput

code423n4:
Auto-generated PR to register the new warden smit_rajput

@smitrajput


SamusElderg:
The user is probably `Smit#3213` but has not yet checked into the `i-want-to-be-a-warden` channel anyway

SamusElderg (approved):

PR #5012, opened 2022-10-29
Register warden asserouad

code423n4:
Auto-generated PR to register the new warden asserouad

@asserouad


SamusElderg:
The user is probably `asserouad67#4496` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #5063, opened 2022-10-29
Register warden Grayson

code423n4:
Auto-generated PR to register the new warden Grayson

@hendry130788


Grayson1307 (commented):
Good

SamusElderg:
The user is probably: `grayson885#5069` but has not yet checked into the `i-want-to-be-a-warden` channel nor does their nickname match the application

PR #5253, opened 2022-11-03
Register warden sagar

code423n4:
Auto-generated PR to register the new warden sagar

@sagar6199


SamusElderg:
The user is probably `Sagar6199#7449` but has not yet checked into the `i-want-to-be-a-warden` channel anyway

PR #5266, opened 2022-11-03
Register warden Amine

code423n4:
Auto-generated PR to register the new warden Amine


SamusElderg:
The user is probably `Amine!!#5196` but has not yet checked into the `i-want-to-be-a-warden` channel anyway

PR #5274, opened 2022-11-04
Register warden Peacher260725

code423n4:
Auto-generated PR to register the new warden Peacher260725

@Peacher260725


0xSorryNotSorry:
The applicant has not shown up in the #i-want-to-be-a-warden channel

SamusElderg:

PR #5305, opened 2022-11-04
Register warden beyay

code423n4:
Auto-generated PR to register the new warden beyay

@beyay


SamusElderg:
The user might be `beyay($POX)#5093` but has not yet checked into the `i-want-to-be-a-warden` channel anyway

PR #5314, opened 2022-11-04
Register warden RomanH

code423n4:
Auto-generated PR to register the new warden RomanH

@RomanH


SamusElderg:
The user is probably `Roman H.#6076` but has not yet checked into the `i-want-to-be-a-warden` channel anyway

PR #5437, opened 2022-11-09
Register warden Cineos

code423n4:
Auto-generated PR to register the new warden Cineos


0xSorryNotSorry:
The applicant has not showed up in #i-wnt-to-be-a-warden channel.

SamusElderg:

PR #5466, opened 2022-11-10
Show end time and start time for active and upcoming contests

captainmangoC4:


captainmangoC4:
We need to show UTC in addition to local timezone.

PR #5801, opened 2022-12-04
Register warden skepticalsec

code423n4:
Auto-generated PR to register the new warden skepticalsec

@skepticalsec


SamusElderg:

SamusElderg:
Cant be pinged publicly in the channel, will leave it open for another day then close the PR if they don't check in

sksb2180:
User checked

PR #5843, opened 2022-12-07
Ignore casing when comparing new username to existing usernames

captainmangoC4:


src/components/RegistrationForm.tsx
```diff
@@ -353,7 +353,8 @@ export default function RegistrationForm({ handles }) {
   const usernameValidator = useCallback(
     (value: string) => {
       const validationErrors: (string | React.ReactNode)[] = [];
-      if (handles.has(value)) {
+      const lowerCaseHandles = handles.map((handle) => handle.toLowerCase());
+      if (lowerCaseHandles.has(value.toLowerCase)) {
```
joshuashort, line 357:
```suggestion
      if (lowerCaseHandles.has(value.toLowerCase())) {
```

joshuashort: This also might be better as handles.find() instead of constructing an entire list of lowercase handles to then search through.

captainmangoC4: Good idea

captainmangoC4: @joshuashort can you take another look?

src/components/RegistrationForm.tsx
```diff
@@ -353,7 +353,11 @@ export default function RegistrationForm({ handles }) {
   const usernameValidator = useCallback(
     (value: string) => {
       const validationErrors: (string | React.ReactNode)[] = [];
-      if (handles.has(value)) {
+      if (
+        [...handles].find(
```
joshuashort, line 357: Oh I thought `handles` was a list here 🙃

```suggestion
        Object.keys(handles).find(
```

PR #5869, opened 2022-12-08
Support Mitigation Review competitions

captainmangoC4:


gatsby-node.esm.js
```diff
@@ -291,6 +299,14 @@ exports.sourceNodes = async ({ actions, getNodes }) => {
...
             : undefined,
       });
+      createNodeField({
+        node,
+        name: `type`,
+        value:
+          dataForCurrentContest.length > 0
+            ? dataForCurrentContest[0].type
+            : undefined,
+      });
```
Simon-Busch, line 309: As the type is defined as `type: String` in the schema, I think we can't return undefined here.

types/finding.ts
```diff
@@ -11,6 +11,8 @@ export interface Finding {
   updatedAt: string;
   issueNumber: number;
   handle: string;
+  isMitigated?: boolean;
+  mitigationOf?: string;
```
Simon-Busch, line 15: Should it be defined as you defined it below ??  
```
 mitigationOf?: {
    newValue: string;
    oldValue: string;
  };
  ```

captainmangoC4: No, the interface definition below is for a request to edit a finding, so it has the previous value and the new value. This is the generic finding interface

Simon-Busch (approved):

PR #5889, opened 2022-12-08
Removing duplicate data

burgertime-c4:
Created from mismatches in the way OSes handle capital letters


joshuashort:
Handled in https://github.com/code-423n4/code423n4.com/pull/5843

PR #5901, opened 2022-12-09
updated submit findings function to handle mititgation submissions

hannahtuttle:


netlify/functions/submit-finding.ts
```diff
@@ -88,10 +88,14 @@ exports.handler = async (event) => {
       }),
     };
   }
+  const isMitigationReport : boolean = mitigationOf ? true : false;
+  let mitigationLabels = ['mitigation-confirmed', `MR-${mitigationOf}`];
+  let mitigationTitle = `Mitigation Confirmed for ${mitigationOf}`;
```
captainmangoC4, line 93: This is a minor thing, but it would be a good idea to make these values "falsey" if `isMitigationReport` is false.

This would make both values `false` if not a mitigation report
```javascript
  const mitigationLabels = isMitigationReport && ['mitigation-confirmed', `MR-${mitigationOf}`];
  const mitigationTitle = isMitigationReport && `Mitigation Confirmed for ${mitigationOf}`;
```

An alternative which would make both values `undefined` if not a mitigation report
```javascript
  const mitigationLabels = isMitigationReport ? ['mitigation-confirmed', `MR-${mitigationOf}`] : undefined;
  const mitigationTitle = isMitigationReport ? `Mitigation Confirmed for ${mitigationOf}` : undefined;
```

netlify/functions/submit-finding.ts
```diff
@@ -248,35 +252,54 @@ exports.handler = async (event) => {
...
+    //if it's a confirmed mitigation then no title will be provided
+    let isConfirmedMitigation = false;
+    if(isMitigationReport && (
+      !body || !title || !risk
+    )){
+      isConfirmedMitigation = true;
+    }else if(
+      isMitigationReport
+    ){
+      labels.push(`bug`)
```
captainmangoC4, line 291: Do you know if there's any reason why we need the bug label? It seems like we should be able to leave it off, but maybe something downstream depends on it.

If we do need the bug label, I can add it back on the frontend.

netlify/functions/submit-finding.ts
```diff
@@ -248,35 +252,54 @@ exports.handler = async (event) => {
...
+      labels.push(`MR-${mitigationOf}`)
```
captainmangoC4, line 292:
```suggestion
      labels.concat(mitigationLabels)
```

netlify/functions/submit-finding.ts
```diff
@@ -248,35 +252,54 @@ exports.handler = async (event) => {
...
+    }
+
     const issueResult = await octokit.request(
       "POST /repos/{owner}/{repo}/issues",
       {
         owner,
         repo,
-        title,
+        title : !isConfirmedMitigation? title : mitigationTitle,
```
captainmangoC4, line 300: if you follow one of my first suggestions for defining mitigationTitle, you can do this:

```javascript
title: mitigationTitle || title,
```

captainmangoC4 (approved):

netlify/functions/submit-finding.ts
hannahtuttle: @captainmangoC4 I just added it in here to keep it consistent with how it has been done, but I don't see why it needs to be here. So I will remove it from here.

PR #6085, opened 2022-12-15
Fix deploy warning

Simon-Busch:


Simon-Busch:
This PR should fix the css warning and a couple of other smaller warnings 

I have checked in the site and didn't find bugs in the css but would love another set of eyes to make sure

burgertime-c4:
I checked everything I could think of locally as well as in the deploy preview. Looks good to me! This is a big change though so we should babysit the site for a bit after merge/deploy

burgertime-c4 (approved):

joshuashort (approved):
ty! 🎉

PR #6203, opened 2022-12-20
Move contest.csv calls to api

hannahtuttle:
This is waiting until the contest data is moved into Moralis.


_data/handles/moksha.json
```diff
@@ -1,4 +0,0 @@
-{
-  "handle": "moksha",
-  "link": ""
-}
```
Simon-Busch, line 4: Is this normal here ?

netlify/functions/leaderboard.ts
```diff
@@ -23,7 +24,7 @@ const getLeaderboardResults = async (
   handle?: string
 ) => {
   // @TODO: also filter by contestId (if provided)
-  const allContests = (await csv().fromFile("_data/contests/contests.csv"))
+  const allContests = (await getApiContestData())
```
Simon-Busch, line 27: See here we are making the API call instead of using the graph for example, so I am not sure we need to feed the graph.

netlify/util/contest-utils.ts
```diff
@@ -1,16 +1,11 @@
-import csv from "csvtojson";
 import { Contest } from "../../types/contest";
+import { getApiContestData } from '../../api/getData';
 
 async function getContest(contestId: number): Promise<Contest | undefined> {
-  const allContests = await csv().fromFile("_data/contests/contests.csv");
+  console.log('get contests');
```
Simon-Busch, line 5:
```suggestion
```

gatsby-node.esm.js
```diff
@@ -273,47 +181,30 @@ exports.onCreateNode = async ({ node, getNode, actions }) => {
...
+    createNode({
+      ...contest,
+      id: createNodeId(`ContestsCsv-${contest.contestid}`),
+      parent: null,
+      children: [],
+      internal: {
+        type: "ContestsCsv",
+        contentDigest: createContentDigest(contest),
+      },
+    });
```
Simon-Busch, line 201: Do we want to create nodes here or do we want to make the API calls in the relevant pages ? as we did for the leaderboard for example?

Simon-Busch:
@joshuashort Would like to have your opinion on [this commit](https://github.com/code-423n4/code423n4.com/pull/6203/commits/455efedf7311c3937a3602ca12061c2e814b1b9d) .

As the contest data will come from the API from now on, I suggest removing the check for contests in `validate.js`. Would it makes sense to you?

joshuashort:
> @joshuashort Would like to have your opinion on [this commit](https://github.com/code-423n4/code423n4.com/pull/6203/commits/455efedf7311c3937a3602ca12061c2e814b1b9d) .
> 
> As the contest data will come from the API from now on, I suggest removing the check for contests in `validate.js`. Would it makes sense to you?

Yeah, agree in principle. *If* we still need to validate the contest csv, that’d have to move to its new home (a _data repo or directly in the API).

But disabling validation here is gonna be needed.

Simon-Busch (approved):

PR #6401, opened 2022-12-31
Register warden raistlen

code423n4:
Auto-generated PR to register the new warden raistlen


SamusElderg:
The user exists but has not yet checked into the `i-want-to-be-a-warden` channel and needs to set their discord nickname in the code4rena channel to exactly `raistlen`

SamusElderg (approved):

PR #6456, opened 2023-01-03
Nighthawk/leaderboard last90days fix

hannahtuttle:


hannahtuttle:
@captainmangoC4 Hey the handle that is deleted here was another case insensitive issue, I made sure that the file matched what was currently in Github, but I'm not sure if this will cause any issues.

joshuashort (approved):

PR #6671, opened 2023-01-11
Register warden yashgoel7291

code423n4:
Auto-generated PR to register the new warden yashgoel7291

@yashgoel72


SamusElderg:
The user is probably `yashgoel20#6381` but has not yet checked into the `i-want-to-be-a-warden` channel nor do they have a matching nickname to the application (needs to be `yashgoel7291`)

SamusElderg:
Havent heard from user, closing PR

PR #6820, opened 2023-01-16
Register warden pastofre

code423n4:
Auto-generated PR to register the new warden pastofre

@pierregvx


SamusElderg:
Might be user `pierre#0362` have asked them to confirm and change nickname to match if so

SamusElderg (approved):

PR #6901, opened 2023-01-19
Build time optimizations

burgertime-c4:
In an effort to reduce build time and prevent memory problems on Netlify, I made a couple changes that made my local builds more efficient. We can see how it performs now on Netlify. 

Changes made: 
- Image optimizing for warden avatars
- Commenting out the syntax highlighter for markdown (used in reports)
  - We likely want to add it back, and it hasn't been removed completely from the codebase. So this felt like a good in-between state for now while we figure out next steps (remove completely or put it back with a different way of utilizing it).


gatsby-config.js
```diff
@@ -78,16 +78,16 @@ const plugins = [
...
+        // {
+        //   resolve: `gatsby-remark-vscode`,
+        //   options: {
+        //     theme: "Dark+ (default dark)",
+        //     languageAliases: {
+        //       solidity: "js",
+        //       sol: "js",
+        //     },
+        //   },
+        // },
```
Simon-Busch, line 90: I am not sure about this.

This was requested by the community

burgertime-c4:
@captainmangoC4 Had the idea to take the syntax highlighting out of the build process, and instead do it on-demand. 🙌

Mango, do you think you could update this PR from here with that change?

burgertime-c4:
I changed the name to remove `wait for approval before merge` since we talked through the future of the highlighter, which is what I wanted confirmation on before merging.

captainmangoC4 (approved):

captainmangoC4:
I brought back syntax highlighting and wasn't able to optimize that, but I also pruned warden files for wardens and teams that have never been active in contests. Hopefully that helps.

PR #7052, opened 2023-01-24
Register warden capriza

code423n4:
Auto-generated PR to register the new warden capriza

@rechlis


SamusElderg:
Can not find a matching user in Discord, will leave the application open for now though
Might be user: `rechlis#5296` see if they check in / reach out and go from there

PR #7055, opened 2023-01-24
Register warden joe

code423n4:
Auto-generated PR to register the new warden joe

@jfrazier-eth


SamusElderg:
Might be user: ` joe#1193` but they haven't checked into the `i-want-to-be-a-warden` channel so probably not, will leave open for now

kartoonjoy (approved):

PR #7058, opened 2023-01-24
Add details field for confirmed mitigations

captainmangoC4:


Simon-Busch (approved):
Seems good to me

burgertime-c4:
I can't see the form without a warden login, but checking the code seems good to me. If you'd like a more thorough review, I could work on that, just let me know!

burgertime-c4 (approved):

PR #7069, opened 2023-01-25
Register warden marvinkruse

code423n4:
Auto-generated PR to register the new warden marvinkruse

@marvinkruse


marvinkruse:
I actually just found out that I have registered before, the account is [pseudorandom](https://github.com/code-423n4/code423n4.com/blob/main/_data/handles/pseudorandom.json). How can I fix this now? I can't push to the PR branch of course, but I guess I would need to update `pseudorandom.json` and rename that to `marvinkruse.json` instead of just creating this new one?

SamusElderg:
> I actually just found out that I have registered before, the account is [pseudorandom](https://github.com/code-423n4/code423n4.com/blob/main/_data/handles/pseudorandom.json). How can I fix this now? I can't push to the PR branch of course, but I guess I would need to update `pseudorandom.json` and rename that to `marvinkruse.json` instead of just creating this new one?

Check in with kartoonjoy on Discord, there is a reply to you there in regard to this 👍

kartoonjoy (approved):

PR #7186, opened 2023-01-30
Register warden looter_

code423n4:
Auto-generated PR to register the new warden looter_

@lootboi


SamusElderg:
The user exists but has not yet checked into the `i-want-to-be-a-warden` channel. Closing PR for now, will re-open once checked in

PR #7201, opened 2023-01-30
[Big PR] Style updates

burgertime-c4:


src/styles/_variables.scss
```diff
@@ -0,0 +1,120 @@
...
+
+$spacing__xxs: calc(#{$spacing__base}/ 4); // 2px
+$spacing__xs: calc(#{$spacing__base}/ 2); // 4px
+$spacing__s: $spacing__base; // 8px
+$spacing__m: calc(#{$spacing__base} * 2); // 16px
+$spacing__l: calc(#{$spacing__base} * 3); // 24px
+$spacing__xl: calc(#{$spacing__base} * 4); // 32px
+$spacing__xxl: calc(#{$spacing__base} * 5); // 40px
+$spacing__xxxl: calc(#{$spacing__base} * 6); // 48px
+
```
natevw, line 46: I like the idea of having spacing variables! I wonder if they could be supplemented with:

1. do you ever use relative spacing (e.g. `1ex` or `0.5em`) for times when you *do* want things like padding to scale with any font size changes? if so maybe a few `$relspacing_` variants might be useful here?
2. do you ever end up "cheating" from time to time with like `margin-top: -1px;` or whatever just to fix up a silly edge case that is defying all troubleshooting? I wonder if something like a `$fudge_pixel: 1px` that you mark usages like `position: relative; left: -3 * $fudge_pixel;` would be useful?

src/styles/_variables.scss
```diff
@@ -0,0 +1,120 @@
...
+$color__n-10: rgba(173, 171, 178, 1);
+$color__n-5: rgba(196, 194, 204, 1);
+$color__n-2: rgba(230, 230, 230, 1);
+
+// Spacing
+// ------------------------------------------------
+
+$spacing__base: 0.5rem; // 8px
+
+$spacing__xxs: calc(#{$spacing__base}/ 4); // 2px
```
natevw, line 38: I was going to ask why these were left as `calc()` but it looks like SASS does [simplify them down](https://sass-lang.com/documentation/values/calculations) when it can so +1 on this syntax. (i.e. thought it might make the every browser redo the calcs but that's not the case.)

It also seems a bit suspicious to be bothering with rems if what you really mean is all the pixels in the comments, but I guess that's kind of the age-old CSS units debates and the idea is that it *could* scale without loss of generality…? https://css-tricks.com/is-it-better-to-use-ems-rems-than-px-for-font-size/

natevw: Oh, actually by that same https://sass-lang.com/documentation/values/calculations link isn't it saying you should be using `calc($spacing__base / 4)` instead of `calc(#{$spacing__base}/ 4)`? So I think this might be something to change so that it **can** pre-calculate all these when it compiles?

src/styles/_variables.scss
```diff
@@ -0,0 +1,120 @@
...
+// Z-index
+// ------------------------------------------------
+
+$z-index__base: 1;
+$z-index__modal: 1000;
+
+// Media queries
+// ------------------------------------------------
+
+@mixin breakpoint($breakpoint) {
```
natevw, line 88: This works, but for future reference this is the kind of thing that a [map](https://sass-lang.com/documentation/values/maps) variable is good for: https://sass-lang.com/documentation/variables#advanced-variable-functions

src/styles/_variables.scss
```diff
@@ -0,0 +1,120 @@
...
+  @if $breakpoint == "xs" {
+    @media (min-width: $breakpoint__xs) {
+      @content;
+    }
+  } @else if $breakpoint == "s" {
```
natevw, line 93: This if/else style works, but for future reference this is the kind of thing that a [map](https://sass-lang.com/documentation/values/maps) variable is good for: https://sass-lang.com/documentation/variables#advanced-variable-functions

src/styles/_variables.scss
burgertime-c4: I love both of these suggestions so much! I think I'll leave 2 for when it's needed, but I'm definitely doing 1 immediately! If you check out `typography.scss` you'll see I hardcoded some em values.

src/styles/_variables.scss
burgertime-c4: whaaaat since when can you use variables without interpolation inside sass's calc? TIL! That's great.

src/styles/base/reset.scss
```diff
@@ -0,0 +1,138 @@
+// css-init by @sir-sharkey https://github.com/sir-sharkey/css-init/blob/main/dist/init.css
+// v 1.2.14
+
+// Modifications:
+// - Removed `color: initial` from `*, :after, :before`
```
natevw, line 5: sounds like there's a couple more modifications that should be included here?

src/styles/base/layout.scss
```diff
@@ -0,0 +1,57 @@
...
+
+// [Tip] For typography applied to `body`, check `base-typography`.
+body {
+  padding: 0 $spacing__l $spacing__m;
+  @include breakpoint("m") {
+    padding: 0 $spacing__l $spacing__l;
+  }
+
+  background: black;
+  color: white !important;
```
natevw, line 24: there's variables for these colors now, right?

src/styles/_variables.scss
```diff
@@ -0,0 +1,120 @@
...
+
+// Transitions
+// ------------------------------------------------
+
+$transition__base: 0.2s ease-in-out;
+
+// Border radius
+// ------------------------------------------------
+
+$border-radius__base: 0.2rem; // 4px
```
natevw, line 75: `$spacing__xs`?

_data/handles/Moksha.json
```diff
@@ -1,5 +0,0 @@
-{
-  "handle": "Moksha",
```
natevw, line 2: is this file part of updating styles?

burgertime-c4: These are weird git cruft because we use user generated data as file names, and different operating systems handle capital letters differently. All things that should not be happening 😅

src/styles/_variables.scss
burgertime-c4: Decided to keep this one as its own thing so we can adjust it separately

sockdrawermoney (approved):
yolo

sockdrawermoney:
Screenshot 2023-03-01 at 5 49 54 PM

sockdrawermoney (approved):
approved with all of my heart

PR #7235, opened 2023-01-31
[Only merge if needed] commenting out syntax highlighting during builds

burgertime-c4:


natevw:
Now that #7582 this will no longer be helpful, the syntax highlighting stuff here is really ± inactive now having been replaced in practice by the copy in the pre-rendering action.

PR #7236, opened 2023-01-31
Register warden markohacker

code423n4:
Auto-generated PR to register the new warden markohacker

@jovanovskmarko


SamusElderg:
The user is probably `markojovanovski#0689` but has not yet checked into the `i-want-to-be-a-warden` channel with their nickname in there set as `markohacker`

SamusElderg:

PR #7277, opened 2023-02-02
Register warden viet-nv

code423n4:
Auto-generated PR to register the new warden viet-nv

@viet-nv


SamusElderg:
The user is probably `viet-nv.bnb#8480` but will need to set local nickname to `viet-nv` and check into the `i-want-to-be-a-warden` channel

SamusElderg:

PR #7278, opened 2023-02-02
Fix leaderboard

Simon-Busch:
- [x] Change netlify function to get the data from graphql tree in `leaderboard.tsx`


hannahtuttle (changes requested):
So the speed on this is great, i noticed that the function was changed from a get to a post. So the leaderboard on the contest pages are not showing up. My guess is that in the contestLayout.tsx you will need to updated the fetch call to the leaderboard to be a post ( I think it's still a get), you could possibly add the grapgQL contest data here too which would speed these individual leaderboard call as well.

src/templates/ContestLayout.tsx
```diff
@@ -118,23 +119,26 @@ const ContestLayout = (props) => {
       }
     })();
   }, [currentUser, contestid, fields]);
-
+  console.log(data.contestsCsv);
```
hannahtuttle, line 122: remove this log

hannahtuttle (approved):
Looks good to me!

PR #7412, opened 2023-02-07
Register warden M1Malikzay

code423n4:
Auto-generated PR to register the new warden M1Malikzay

@sebghatullahmalikzay


SamusElderg:
The user is probably `sebghatullah#6745` but has not yet checked into the `i-want-to-be-a-warden` channel with matching nickname<>C4Username yet

SamusElderg:

Issue #7424, opened 2023-02-07
Optimization: load ReportLayout content directly from HTML


natevw:
The site build times are problematic; there are some tricks like #7235 that can speed up the build in a "break glass" situation but at the expense of readability/polish in the rendered contest reports.

The thing is that we have hundreds of contest reports that really rarely should ever have to be re-rendered. Most haven't been edited for months. Gatsby is supposed to magically handle this sort of thing, but for reasons we can't always rely on cached pre-built content from prior runs. We need to be able to do an enitre clean build without hitting Netlify timeouts.

The plan is that we have Gatsby generate the contest reports portion of the site only needing to fill in those pages with pre-processed HTML, rather than churning the whole https://github.com/code-423n4/code423n4.com/tree/a9d0324/_data/reports folder full of .md files through [a whole big stack of processing](https://github.com/code-423n4/code423n4.com/blob/ff570ec/gatsby-config.js#L43-L90) on a clean build.

natevw:
To accomplish this I think the code at https://github.com/code-423n4/code423n4.com/blob/fe6bcad/src/pages/reports/%7BMarkdownRemark.fields__slug%7D.js can remain largely untouched — it's mainly that the source data for

```
<div
  className="report-contents"
  dangerouslySetInnerHTML={{ __html: DOMPurify.sanitize(page.html) }}
/>
```

needs to be "ready baked" through a simpler transform. According to https://github.com/gatsbyjs/gatsby/issues/9721 there is no built-in transform at least back at that time. It looks like https://www.gatsbyjs.com/plugins/gatsby-transformer-rehype/ is able to source HTML content into Gatsby but is probably overkill.

Unless I find something or it turns out accessible within Gatsby's node graph already, my plan is to develop a small custom transformer plugin to meet our needs, such that it really turns into simply exposing the file content read directly as string data to the `ReportPageTemplate` component.

natevw:
Random thought at some point when I was off my laptop: there may be some TOC and other related variables that *also* come out of the Markdown rendering process? but regardless, thinking about this helped me realize that an even more expedient path than convincing Gatsby to slurp up *.html file content into variables would be to simply put wrap that HTML inside *.json files…

so `_data/reports/2023-fake_contest.json` would be at minimum just a raw JSON-escaped string but probably at least wrapped in an actual object:

```
{
  "teh_rendered_html": "…"
}
```

which then could be extended to include solutions for a few other lingering open questions e.g.:

```
{
[...]

PR #7581, opened 2023-02-16
Drop gatsby-plugin-mdx plugin

natevw:
I noticed this "extra" Markdown plugin while working on the "real" Markdown processing for #7424. It's apparently for https://mdxjs.com/ which is this like React-in-Markdown thing?!

But afaict this is not needed:

* there are no .mdx files *in* the repo
* we would not really want to process MDX from *outside* the repo
* it does not seem to be in use and the site builds fine without the plugin

So unless someone knows why it's there or when it would be used I think we can drop it?


SamusElderg:
Closing stale PR

PR #7611, opened 2023-02-17
Tidy up some getApiContestData loose ends

natevw:
This came out of some troubleshooting attempts lately when the fetch doesn't always work. Specific things addressed:

1. The extension was `.ts` but usage of the file from gatsby-node.esm.js breaks if Typescript syntax is actually used. So I changed the file extension to `.js`.
2. …but I did still annotate the types, using various [JSDoc syntax](https://jsdoc.app/tags-async.html) that VSCode supports (but n.b. doesn't by default lint and warn as aggressively as it does with actual TypeScript!)
3. Significantly changed how errors are handled! Now they just "bubble up" rather than getting quietly logged and then resolving to `undefined`. (All the current callers crash-and-burn if the original catch block is triggered and returns undefined; returning an empty array in that case still tends to lead to trouble down the road, and seems wrong anyway. So IMO it's better to pass the original error up, instead of waiting for some sort of harder-to-troubleshoot error happen a few lines later.)
4. Also, the helper no longer tries to resolve with parsed JSON in cases where the API server returned some sort of unexpected (i.e. non-`200`) status code. Instead it does log the response body as text to aid troubleshooting, and rejects with that considered an error.

So fairly significant changes but afaict should leave things generally in the same situation, only now with more specific fetch-related errors being thrown instead of random ones of trying to use Array methods on an `undefined` contests variable.

(Also fwiw I did dabble with returning an empty array `[]` of contests on error. That lets things get a little farther but later down the road other at least a few things still get pretty upset and throw their own exceptions if there's not e.g. a contest to sort reports list on and such. Generally I don't think that's the right approach anyway… if contests can't be fetched that seems like time to complain rather than pretending there are just 0 contests.)


natevw:
@hannahtuttle What do you think of these changes? I got a bit carried away but the main goal here is so that if contests fail to fetch we get a clearer `fetch`/parse-related error early saying something like timeout/bad URL/bad status/bad data instead of just the coincidental `.forEach` explosion on the missing array.

I'd like to merge this before the next API timeout or whatnot happens if it looks reasonable to you.

hannahtuttle (approved):

hannahtuttle:
@natevw hey I approved these changes, would you like me to merge this?

PR #7636, opened 2023-02-17
Avoid generating extra copy of contest report HTML

natevw:
Each contest report has been getting generated in two places:

* as `ReportLayout` pages under `/reports/{MarkdownRemark.fields__slug}` (intended)
* as `DefaultLayout` pages at the top level `/{MarkdownRemark.fields__slug}` (likely not!)

You can see this now on the live site e.g. the broken `https[:]//code4rena.com/2022-12-tigris/` vs. the actual `https[:]//code4rena.com/reports/2022-12-tigris/`.

I don't think this would have added a huge amount of extra build time (since the actual markdown needs to be processed into the graph regardless) but certainly is some extra work that wouldn't help.

This change is admittedly kind of hacky but we should be able to drop it entirely after #7582 is wrapped and merged and the only actual MarkdownRemark nodes left will be the `_data/pages/*.md` ones instead of all the `_data/reports/*.md` ones too.


natevw:
No longer needed, the changes in #7582 also avoid the duplicate reports through another way.

PR #7672, opened 2023-02-20
Try turning off source map generation

natevw:
This adds `gatsby-plugin-no-sourcemaps` as recommended by https://v3.gatsbyjs.com/docs/how-to/performance/resolving-out-of-memory-issues/#3-reducing-crashes-in-generating-javascript-bundles--serializing-html-pages for improving both memory and build times. Keeping in mind:

> The side effect is that this will make any errors happening in production significantly harder to debug by obscuring the stacktrace, so if your team regularly inspects production errors (perhaps using eg Sentry, Rollbar, or comparable tools), you may want to consider other options.

Notes/caveats:

- [ ] I didn't notice a huge clear benefit to this, but otoh haven't done extensive A/B testing yet either
- [ ] I haven't included any commit changing the `yarn.lock` file which probably needs to happen with a package change too?
- [ ] seems like there was one other thing I wanted to mention? hate that feeling!


SamusElderg:
Closing stale PR

PR #7698, opened 2023-02-21
fix typo

nate-dag:
there's a Markdown typo in the readme


natevw:
This is on the radar and actually already fixed in #7579 just hasn't merged yet.

PR #7708, opened 2023-02-22
Register warden hamzaarc

code423n4:
Auto-generated PR to register the new warden hamzaarc


SamusElderg:
The user is probably `hamzaarchid#5292` but has not yet checked into the `i-want-to-be-a-warden` channel and will need to set their local nickname (not their global handle) in the C4 server to exactly: `hamzaarc`

SamusElderg:

PR #7761, opened 2023-02-24
Register warden dingo2077

code423n4:
Auto-generated PR to register the new warden dingo2077

@dingo2077


SamusElderg:
Have contacted user to change their nickname to match application, but they will likely try re-register with a different username selection

SamusElderg (approved):

PR #7840, opened 2023-02-28
Nighthawk/rebranderino fixes

hannahtuttle:
- [x] remove emojis from leaderboard
- [x] weird border color on contest mobile view
- [x] weird color border on the contest page live banner
- [x] leaderboard - when you sort down, flip the rank number too so someone isn’t highlighted like first least medium findings
- [ ] bug when leaderboard is sorting sometime the usd values are sorted and the dollar amounts are sorted correctly but the change on that dollar amount is sorted incorrectly so the ranks are incorrect when the values are differenc based on the change. EX:  $55.20 is sorted as Rank 100 and  then $55.90 is ranked as 101.

### DIFF
--- a/src/components/LeaderboardTableReduced.js
+++ b/src/components/LeaderboardTableReduced.js
@@ -9,7 +9,32 @@ const LeaderboardTableReduced = ({ results, isLoading, reduced }) => {
       {
         Header: "#",
         Cell: (props) => {
-          return <p>{props.flatRows.indexOf(props.row) + 1}</p>;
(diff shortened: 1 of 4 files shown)


hannahtuttle:
@burgertime-c4 I wasn't sure what you meant by a weird color on contests but I did see the stranger border on the mobile view for contest so here is the commit for those changes https://github.com/code-423n4/code423n4.com/pull/7840/commits/020ef87f94877af7e88a70fa8d48d96dcc0aa7b3 is this what you meant?

PR #7885, opened 2023-03-02
meta content update: default description, homepage description

burgertime-c4:


Simon-Busch (approved):

src/templates/WrapperLayout.js
```diff
@@ -22,7 +22,15 @@ const WrapperLayout = (props) => {
         <meta charset="utf-8" />
         <meta httpEquiv="X-UA-Compatible" content="chrome-1" />
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-        <meta name="description" content={pageDescription} />
+
+        {pageDescription ? (
+          <meta name="description" content={pageDescription} />
+        ) : (
```
natevw, line 28: for future situations like this, it can be cleaner to do something like:

```
const DEFAULT_PAGE_DESCRIPTION = "Top auditors etc. etc. … …";

/* above somewhere in page scope, enabling the below technique */

<meta name="description" content={pageDescription || DEFAULT_PAGE_DESCRIPTION} />
```

PR #7987, opened 2023-03-07
Fix fetching of findings

captainmangoC4:
Some wardens have reported not being able to see their findings after submitting them. This was due to improper handling of usernames with hyphens in them. This PR fixes that bug and also fetches team findings and user findings in parallel to improve performance and reduce timeouts.


netlify/util/github-utils.ts
```diff
@@ -195,10 +195,11 @@ async function getSubmittedFindingsFromFolder(
           })
           .map((f) => {
             const [key, ext] = f.name.split(".");
-            const [handle, issueNumber] = key.split("-");
+            const splitFileName = key.split("-");
+            const issueNumber = splitFileName.pop();
```
natevw, line 199: This is a small function so it's easy to connect the dots, but I would rather see handle explicitly assigned here too and then never use `splitFileName` again after that, since already after this pop it's no longer actually the  split file name anymore. (I sometimes use an underscore naming convention like `_splitFileName` to indicate sort of a "temporary"/"messy" variable like this one but dunno how common or following any code standard that is.)

captainmangoC4: Thanks for the feedback! Implemented.

natevw:
@captainmangoC4 I tidied up the title of this PR a bit but I'm not sure what I'm reviewing here:

> No description provided

What's the context and/or issue is this addressing? And can you summarize how this PR changes things?

captainmangoC4:
Yep @natevw it's a fix for failing to fetch contest findings for some wardens. I requested your review because you had left code comments previously.

netlify/functions/manage-findings.ts
```diff
@@ -90,44 +90,85 @@ async function getFinding(
...
+  const promises: Promise<void>[] = [];
+  for (const teamName of teamNames) {
+    promises.push(
+      getSingleTeamFindings(teamName, octokit, contest).then((teamFindings) => {
+        wardenFindingsForContest.teams[teamFindings.teamName] =
+          teamFindings.findings;
+      })
+    );
+  }
+  return await Promise.all(promises);
```
natevw, line 108: This is okay but might be better more just inline:

```
return await Promise.all(
  teamNames.map((teamName) => getSingleTeamFindings(teamName, octokit, contest).then(/* etc. */))
);
```

…but I'm seeing something potentially more notable as I look more closely, will add a comment above next.

netlify/functions/manage-findings.ts
```diff
@@ -90,44 +90,85 @@ async function getFinding(
...
+  teamNames: string[],
+  octokit: Octokit,
+  contest: Contest,
+  wardenFindingsForContest: WardenFindingsForContest
+): Promise<void[]> {
+  const promises: Promise<void>[] = [];
+  for (const teamName of teamNames) {
+    promises.push(
+      getSingleTeamFindings(teamName, octokit, contest).then((teamFindings) => {
+        wardenFindingsForContest.teams[teamFindings.teamName] =
```
natevw, line 103: Haven't gotten to the calling code yet but assuming there's a really compelling reason to do this as a side effect then this should be flagged in a comment so it stands out!

Otherwise refactor to more like:

```
async function getTeamsFindings(
  teamNames: string[], octokit: Octokit, contest: Contest
): Promise<WardenFindingsForContest>
```

natevw: Ah, actually there's two directions here I'm not sure which one you prefer but should probably pick one:

```
let collection;
for (const x of things) {
   const data = await someFetchRelatedTo(x)
   collection[data.thing] = whatever;
}
return collection;
```

Or otherwise:

```
return await Promise.all(
  things.map(gatherThem)
).then((arr) => arr.reduce((collection, data) => {
  collection[data.thing] = whatever
  return collection;
}, {});
```
if you like `reduce` but probably more clearly:

```
return await Promise.all(
  things.map(gatherThem)
).then((arr) => {
  let collection = {};
  arr.forEach((data) => {
    collection[data.thing] = whatever
  });
  return collection;
```

But either way get rid of giving a name to `promises` especially when they don't resolve to anything useful.

natevw: **UDPATE**: aaaaah forgot the whole point here based on one of the commit messages of this PR, is to make it parallel!


So I misunderstood some code above, but I still maintain that this generically named `promises` is a code smell and we should establish a better pattern for this.

Problem: we're waiting to process an array of input in parallel, using promises
Context: when we [do, or alternately do NOT] know that the array length is small

The solution depends on the expected array size. If you know it's small you can maybe do it ± this way just with a little clearer code. If you know it could be big then you probably do need a better-enapsulated strategy from the start.

netlify/functions/manage-findings.ts
```diff
@@ -210,32 +251,36 @@ async function editFinding(
...
+  let oldRiskCode = "";
+  let newRiskCode = "";
+  if (!data.mitigationOf) {
     oldRiskCode = getRiskCodeFromGithubLabel(data.risk.oldValue);
     newRiskCode = getRiskCodeFromGithubLabel(data.risk.newValue);
-  }else{
-    oldRiskCode = data.risk.oldValue ? getRiskCodeFromGithubLabel(data.risk.oldValue) : '';
-    newRiskCode = data.risk.newValue ? getRiskCodeFromGithubLabel(data.risk.newValue) : '';
+  } else {
+    oldRiskCode = data.risk.oldValue
```
natevw, line 268: I thought we had already caught each repo up with this sort of linting?

natevw (approved):
I think we need to regroup sometime around our "doing things in parallel" code and I've left a bit of a mess of commentary re. that here **but** my understanding is this is something of a hotfix.

AFAICT this code is correct logic-wise and so I don't want to block it on more stylistic concerns that aren't really specific to the urgent issue here.

netlify/functions/manage-findings.ts
captainmangoC4: Maybe not this one? I don't know. I just auto lint on save.

netlify/functions/manage-findings.ts
captainmangoC4: In probably 99% of cases, the array size is going to be 0 or 1. There are few wardens on multiple teams. Given that context, maybe it's fine to just fetch team findings one at a time, but in parallel with warden findings? Also, moving this function to the api will speed things up because we won't have to wait for a response from the api to get contest data. Ideally that move will happen soon.

netlify/functions/manage-findings.ts
natevw: yep, made the PR harder to review but not a blocker.

netlify/functions/manage-findings.ts
natevw: Nah, my assumption in approving this PR was just that. I think what you have will work for now and I'll file a followup task for myself to propose a longer-term pattern to circle back and clean up the parallel code stuff that's not just here 
but in iirc already other places too.

PR #8012, opened 2023-03-08
Bg/form validation

burgertime-c4:


captainmangoC4 (approved):
Looks great! Just clarify the login method as being username not email.

PR #8233, opened 2023-03-14
Register warden ToonVH

code423n4:
Auto-generated PR to register the new warden ToonVH


SamusElderg:

ToonVanHove:
> The user exists but has not yet checked into the `i-want-to-be-a-warden` channel

Should be fixed! :)

SamusElderg (approved):

PR #8263, opened 2023-03-15
updated gatsby node esm to uses contest api data

hannahtuttle:


natevw (changes requested):
+1 on changing the fields like `contest_id` to be consistent

but let's roll back the changes to displaying `award` and just keep that as a single display string from the API.

natevw (approved):
The actual code changes now look exactly just the small tweak expected here, so I'm approving them!

That said, the unrelated file changes in `_data/handles` are a little concerning? Is that just something that GH is confused about with all the branch merging and stuff, or are we accidentally undoing a recent CA change or something there?

Simon-Busch (approved):

PR #8283, opened 2023-03-16
Register warden rohrschach

code423n4:
Auto-generated PR to register the new warden rohrschach

@promachina


SamusElderg:
The user is probably `promachina#6598` but has not yet checked into the `i-want-to-be-a-warden` channel with a matching nickname yet

sksb2180 (approved):

PR #8285, opened 2023-03-16
Add linter + pre-commit hook

Simon-Busch:


.prettierignore
```diff
@@ -0,0 +1,8 @@
+# Ignore artifacts:
+.netlify
+_data
```
natevw, line 3: I think also `public` (Gatsby output folder) should get added to this list too?

package.json
```diff
@@ -106,12 +108,18 @@
   "scripts": {
     "build": "gatsby build",
     "develop": "gatsby develop",
-    "format": "prettier --write \"**/*.{js,jsx,ts,tsx,json,md}\"",
+    "format": "prettier --write \"*/**/*.{js,jsx,ts,tsx,json,css,scss}\"",
```
natevw, line 111: +1 to changing which extensions

but why change from `**/` to `*/**/` here? by adding a level it stops checking the top-level files which I don't think is intended?

package.json
```diff
@@ -106,12 +108,18 @@
...
     "serve": "gatsby serve",
     "clean": "gatsby clean",
     "test": "echo \"Write tests! -> https://gatsby.dev/unit-testing\" && exit 1",
-    "validate": "node bin/validate.js"
+    "validate": "node bin/validate.js",
+    "prepare": "husky install"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "prettier --write \"*/**/*.{js,jsx,ts,tsx,json,css,scss}\""
```
natevw, line 121: same here as above, I think this should be:

```
"prettier --write \"**/*.{js,jsx,ts,tsx,json,css,scss}\""
```

i.e. without the extra `*/` at start of the pattern?

natevw (changes requested):
Overall looks good, thanks! Just a couple tweaks I think to the ignore and to the file patterns it checks.

.prettierignore
Simon-Busch: Exactly, I'll add it right away

natevw (approved):

PR #8290, opened 2023-03-16
Register warden hiennhi

code423n4:
Auto-generated PR to register the new warden hiennhi

@conghien


SamusElderg:
The user is probably `hiennhi959701#9299` but has not yet checked into the `i-want-to-be-a-warden` channel with a matching nickname

PR #8346, opened 2023-03-17
Register warden eric

code423n4:
Auto-generated PR to register the new warden eric

@Ritleboufon


SamusElderg:
The user is probably `ritleboufon#4182` but has not yet checked into the `i-want-to-be-a-warden` channel

Issue #8509, opened 2023-03-23
[BUG]: Fix discord link in "For Wardens" homepage


0xdevant:
## Describe the bug
In "For Wardens" homepage's "Are you ready?" section, the discord link of "Get started" being relative link leading to a 404 error

## To Reproduce
1. At the homepage, which is code423n4.com or code4rena.com
2. Click on "For Wardens" in the first section of the homepage
3. Scroll down to "Are you ready?..."
4. Click on "Get started"

## Expected behavior
Expected to lead to the discord server of Code4rena which should be linked as https://discord.gg/code4rena

## Screenshots
The link becomes "https://code4rena.com/https://discord.gg/code4rena" as shown in the screenshot
<img width="1045" alt="Screenshot 2023-03-23 at 10 33 35 AM" src="https://user-images.githubusercontent.com/26415019/227085328-0b18daed-3572-44e6-9902-f1d6b58d21da.png">

## Desktop:
 - OS: macOS
 - Browser: Brave

## Smartphone (please complete the following information):
Same for desktop

0xdevant:
Please only review the latest commit ffaa201, thank you.

PR #8510, opened 2023-03-23
#8509 Fix discord link for bottomCTA in warden videMode of homepage

0xdevant:
As mentioned in #8509, the discord link should be fixed as an absolute link instead of relative link to solve the 404 error.


Simon-Busch (approved):

0xdevant:
> Good catch, thank you @0xdevant

Anytime!

PR #8635, opened 2023-03-28
Inactivating Team Secureverse

liveactionllama:


_data/handles/Secureverse.json
```diff
@@ -16,4 +12,5 @@
     }
   ],
   "image": "./avatars/Secureverse.png"
```
natevw, line 14: needs to be a comma between this line and the next this otherwise it can't be parsed as JSON

PR #8739, opened 2023-03-30
Nighthawk/move findings data

hannahtuttle:


netlify/util/getFindingsData.ts
natevw, line 1: There was a reason I needed to make the contest version of this be `getContestsData.js` because iirc something uses it as (modern module-friendly) **JavaScript** rather than actual TypeScript.

Is that true for this too? If so then this should be renamed to `.js` too.

netlify/util/getFindingsData.ts
```diff
@@ -0,0 +1,22 @@
+import fetch from "node-fetch";
+import { AwardFinding } from "../../types/finding";
```
natevw, line 2: assuming plain JS is needed, this line can become:

```
/**
 * @typedef {import('../../types/finding').AwardFinding} AwardFinding
 */
 ```

File: types/finding.ts [hunk 2]
```diff
@@ -74,3 +74,17 @@ export interface FindingDeleteRequest {
...
+  handle: string;
+  finding: string;
+  risk: string;
+  score: number;
+  pie: number;
+  split: number;
+  slice: number;
+  award: number;
+  awardCoin: string;
+  awardUSD: number;
```
natevw, line 89: I'm assuming these all preserve backwards compatibility with the CSV/GraphQL version, right? I think that's the best way to leave it for now.

netlify/functions/resources.ts
```diff
@@ -1,5 +1,6 @@
 import { builder, Handler } from "@netlify/functions";
 import { getApiContestData } from "../util/getContestsData";
+import { getApiFindingsData } from "../util/getFindingsData";
 import { readFile } from "node:fs/promises";
```
natevw, line 4: I think this `readFile` import can be dropped now right?

natevw (commented):
Overall looks good to me, but maybe get another set of eyes?

* I did **not** review the `functions/leaderboard.ts` changes **or** the rest of the repo to make sure nothing was using it on the Gatsby/GraphQL side.
* I did have some feedback, minor changes requested if time but no blockers.

netlify/util/getFindingsData.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @returns {Promise<[AwardFinding]>}
+ * @throws Various errors if data cannot be fetched or processed.
+ */
+const getApiFindingsData = async () => {
+  const res = await fetch(`${process.env.C4_API_URL}/api/v0/getFindings`, {
+    method: "POST",
+    body: JSON.stringify({ token: process.env.C4_API_TOKEN }),
+  });
+  if (res.status !== 200) {
+    console.log(await res.text());
```
Simon-Busch, line 15:
```suggestion
```

netlify/util/getFindingsData.ts
```diff
@@ -0,0 +1,22 @@
+import fetch from "node-fetch";
+import { AwardFinding } from "../../types/finding";
+
+/**
+ * @async
+ * @returns {Promise<[AwardFinding]>}
+ * @throws Various errors if data cannot be fetched or processed.
+ */
+const getApiFindingsData = async () => {
```
Simon-Busch, line 9:
```suggestion
const getApiFindingsData = async (): Promise<AwardFinding[]>=> {
```

bin/validate.js
```diff
@@ -153,56 +152,52 @@ async function validateOrganizations() {
...
+//   console.log("✅  Findings validation passed!");
+// }
 
 (async () => {
   try {
     await validateHandles();
     await validateTeams();
     await validateOrganizations();
-    await validateFindings();
+    // await validateFindings();
```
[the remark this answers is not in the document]
captainmangoC4: It looks like this validation is to make sure that there are wardens and contests to match each finding. I think we can move this to the api

netlify/functions/leaderboard.ts
```diff
@@ -39,28 +40,19 @@ const getLeaderboardResults = async (
...
+    filteredContests.some((contest) => contest.contestid === finding.contest)
+  );
 
   // get deduplicated handles from findings
   const allHandles = Array.from(
-    new Set(allFindings.map((finding) => finding.handle))
+    new Set(allFindings.map((finding: AwardFinding) => finding.handle))
   )
-    .map((handle) => getWardenInfo(handle))
+    .map((handle: any) => getWardenInfo(handle))
```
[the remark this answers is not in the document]
captainmangoC4: I recommend type `Username` from types/shared.ts

netlify/util/getFindingsData.ts
captainmangoC4: I would actually add the result of `res.text()` to the error message you're throwing. We don't want to obfuscate the error and lose the ability to track what went wrong.

netlify/util/getFindingsData.ts
```diff
@@ -0,0 +1,22 @@
+import fetch from "node-fetch";
+import { AwardFinding } from "../../types/finding";
+
+/**
+ * @async
+ * @returns {Promise<[AwardFinding]>}
+ * @throws Various errors if data cannot be fetched or processed.
```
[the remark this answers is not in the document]
natevw: The original was valid JSDoc, and I don't think this suggestion adds much?

`@returns {Promise<[AwardFinding]>}` already says "returns a Promise containing an array of AwardFinding objects" to both human **and** the IDE.

netlify/util/getFindingsData.ts
natevw: See my note at https://github.com/code-423n4/code423n4.com/pull/8739#discussion_r1156508297 — I'm not sure we can actually use TypeScript here.

If that's correct the file should be renamed to `.js` and left with only the JSDoc annotations in comments.

netlify/util/getFindingsData.ts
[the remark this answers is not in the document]
Simon-Busch: Well nothing obligatory, I think it's clearer, but it's a personnal preference

bin/validate.js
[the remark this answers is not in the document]
hannahtuttle: I'll make a new issue for this to add it to the api.

types/finding.ts
hannahtuttle: Yes, i tried to make sure that for right now it's the same type of data that was used before to make sure it was backwards compatible.

netlify/functions/leaderboard.ts
```diff
@@ -39,25 +40,19 @@ const getLeaderboardResults = async (
     .filter((contest) => !contestId || Number(contestId) === contest.contestid);
```
natevw, line 40: GH won't let me select the *actual* line of code since it's not really this PR, but might be nice to give `const filteredContests` a proper type while editing this code — but only if it's a relatively easy lift and doesn't spiral out into a more complicated PR to review!

netlify/functions/leaderboard.ts
```diff
@@ -39,25 +40,19 @@ const getLeaderboardResults = async (
...
 
   // get deduplicated handles from findings
   const allHandles = Array.from(
-    new Set(allFindings.map((finding) => finding.handle))
+    new Set(allFindings.map((finding: AwardFinding) => finding.handle))
   )
-    .map((handle) => getWardenInfo(handle))
-    .filter((handle) => handle.showOnLeaderboard !== false);
-
+    .map((handle: string) => getWardenInfo(handle))
```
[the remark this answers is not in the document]
natevw: n/m the existing code was already doing this. let's focus on landing these changes and moving on to other API work, rather than optimizing older code right now :-)

netlify/functions/leaderboard.ts
```diff
@@ -39,25 +40,19 @@ const getLeaderboardResults = async (
...
-      return finding;
-    })
-    .filter((finding) =>
-      filteredContests.some(
-        (contest) => contest.contestid.toString() === finding.contest
-      )
-    );
+  const allFindings = (await getApiFindingsData()).filter((finding) =>
+    filteredContests.some((contest) => contest.contestid === finding.contest)
+  );
```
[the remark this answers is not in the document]
natevw: n/m, same sentiment as below — the new code isn't any worse than the old code so it's a +1 from me given the current goals!

natevw (commented):
Looks good to me but going to try actually testing it all locally before formally approving! (Some random style comments but let's just leave it as-is since after I wrote them up realized you were basically just migrating the existing code that's already in production!)

natevw (approved):
Tested and confirmed the leaderboard is working for me locally with this PR.

natevw (changes requested):
Noticing there might be some <http://localhost:8888/community-resources/findings.csv> changes vs. the real <https://code4rena.com/community-resources/findings.csv> — tbd.

netlify/functions/resources.ts
```diff
@@ -68,13 +68,45 @@ const resourcesHandler: Handler = async (event, context) => {
...
+        "awardUSD",
+      ],
+      //contest,handle,finding,risk,score,pie,split,slice,award,awardCoin,awardUSD
+      ...findingsData.map((finding) => {
+        return [
+          "" + finding.contest,
+          finding.handle,
+          finding.finding,
+          finding.risk,
+          "" + finding.score,
```
natevw, line 93: to match the current findings, this needs to generate a `""` string when `finding.score === null`

screenshot of two CSVs open in Numbers.app highlighting column value differences

netlify/functions/resources.ts
```diff
@@ -68,13 +68,45 @@ const resourcesHandler: Handler = async (event, context) => {
...
+          "" + finding.pie,
+          "" + finding.split,
+          "" + finding.slice,
+          "" + finding.award,
+          finding.awardCoin,
+          "" + finding.awardUSD,
```
natevw, line 99: `award` and `awardUSD` are getting truncated down to nearest integer when I test locally — unclear if that's in this code or something happening over on the API side.

Another numbers screenshot with CSV differences highlighted

natevw (changes requested):

netlify/functions/resources.ts
natevw: Okay see https://github.com/code-423n4/api.code4rena.com/pull/534/files#r1169384518 for half of this fix and I'll make a new code suggestion for this side.

types/finding.ts
```diff
@@ -83,6 +83,20 @@ export interface FindingDeleteRequest {
   emailAddresses: string[];
 }
 
+export interface AwardFinding {
+  contest: number;
+  handle: string;
+  finding: string;
+  risk: string;
+  score: number;
```
natevw, line 91: To support the historical data set this will need to be:

```suggestion
  score: number | null;
```

natevw (changes requested):

netlify/functions/resources.ts
natevw:
```suggestion
          (finding.score === null) ? "" : ("" + finding.score),
```

netlify/functions/resources.ts
natevw, line 99: I think after https://github.com/code-423n4/api.code4rena.com/pull/534#discussion_r1169384518 most of these should be good with the default rounding but let's tweak the final column to mostly match current formatting:

```suggestion
          "" + finding.pie,
          finding.split.toPrecision(6),
          "" + finding.slice,
          "" + finding.award,
          finding.awardCoin,
          finding.awardUSD.toPrecision(2),
```

This will still change some of the current `0` cells to become `0.00` but I think that's reasonable and shouldn't cause any backwards-compatibility trouble!

natevw, line 99: I think after https://github.com/code-423n4/api.code4rena.com/pull/534#discussion_r1169384518 most of these should be good with the default rounding but let's tweak the final column to mostly match current formatting:

```suggestion
          "" + finding.pie,
          finding.split.toPrecision(6),
          "" + finding.slice,
          "" + finding.award,
          finding.awardCoin,
          finding.awardUSD.toPrecision(2),
```

This will still change some of the current `0` cells to become `0.00` but I think that's reasonable and shouldn't cause any backwards-compatibility trouble!

hannahtuttle: To follow up, this was fixed with those api changes.

hannahtuttle: This change was formatting the award in USD to look like `1.5E4` instead of keeping it more formatted like you would expect USD. It did format 0 to `0.0`, so i left it the way it was so we don't have that unexpected formatting.

netlify/functions/leaderboard.ts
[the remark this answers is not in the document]
hannahtuttle: Yah iwas trying to change as little code in this repo as possible to keep from causing any issues.

netlify/functions/leaderboard.ts
hannahtuttle: This was an easy lift, there was already a contest type that had a few more things on it then I needed to I just made a simple contest type and extended out to keep the other contest type the same.

types/contest.ts
```diff
@@ -1,7 +1,7 @@
 import { DateString, RelativeURL } from "./shared";
 
 // all values are strings because the source is a CSV file -- changed getting values from api instread of contest.csv
-export interface Contest {
+export interface DBContest {
```
natevw, line 4: won't this break imports from other files, like manage-findings\.ts and others?

search in VSCode for imports from types/contest showing other uses of older Contest name



unless I'm missing something lets keep this as:

```suggestion
export interface Contest {
```

netlify/functions/resources.ts
```diff
@@ -68,13 +68,45 @@ const resourcesHandler: Handler = async (event, context) => {
...
+        "awardUSD",
+      ],
+      //contest,handle,finding,risk,score,pie,split,slice,award,awardCoin,awardUSD
+      ...findingsData.map((finding) => {
+        return [
+          "" + finding.contest,
+          finding.handle,
+          finding.finding,
+          finding.risk,
+          finding.score === null ? "" : "" + finding.score,
```
natevw, line 93: ok, I can never remember the operator precedence of `?:` but looks like this works:

testing the logic with findings score null vs. zero and getting intended results

still might feel "safer" as:

```suggestion
          finding.score === null ? "" : ("" + finding.score),
```

but up to you.

natevw (changes requested):
[still reviewing but one change needed I think re. not renaming the DBContest import!]

netlify/functions/leaderboard.ts
```diff
@@ -1,7 +1,9 @@
 import { differenceInDays, getYear } from "date-fns";
 import fs, { readFileSync } from "fs";
+import { AwardFinding } from "../../types/finding";
 import { getApiContestData } from "../util/getContestsData";
-import csv from "csvtojson";
+import { getApiFindingsData } from "../util/getFindingsData";
+import { DBContest } from "../../types/contest";
```
natevw, line 6:
```suggestion
import { Contest } from "../../types/contest";
```

netlify/functions/leaderboard.ts
```diff
@@ -34,30 +36,25 @@ const getLeaderboardResults = async (
 
   const filteredContests = allContests
     .filter(
-      (contest) => withinTimeframe(contest, contestRange) || !contestRange
+      (contest: DBContest) =>
```
natevw, line 39:
```suggestion
      (contest: Contest) =>
```

netlify/functions/leaderboard.ts
```diff
@@ -34,30 +36,25 @@ const getLeaderboardResults = async (
...
+        withinTimeframe(contest, contestRange) || !contestRange
```
natevw, line 40: wouldn't this work too, and avoid calling the helper function at all when there's a missing range?

```suggestion
        !contestRange || withinTimeframe(contest, contestRange)
```

but maybe there's a subtlety to the original logic that I'm missing where you would want to try check it?

types/contest.ts
hannahtuttle: No I left the contest as an interface, Contest extends from the DBContest to add what is missing and I didn't change any of the types in the interface.

netlify/functions/resources.ts
natevw: [or the linter if it's `prettier` that wants it the current way!]

types/contest.ts
natevw: sorry missed that you sitll have `export interface Contest` below.

natevw (approved):
The types change was fine, I had missed a detail lower down in the file.

PR #8744, opened 2023-03-30
Fix wardens not being able to see their submitted findings

captainmangoC4:
some warden's can't see/manage their findings in the last hours of a contest. ([this fire](https://discord.com/channels/810916927919620096/1072652108555354152/1091064648956129321))
Octokit was not fetching all warden json files because there were so many submissions.
This fixes that bug, plus adds some performance improvements.

 - [x] test whether downloading the zip of the repo content returns more files than octokit.request
 - [x] implement octokit.paginate instead of graphql pagination for performance improvements
 - [x] make calls to fetch finding json files and issues in parallel, then merge after both succeed to improve performance
 - [x] make sure to only fetch all json files and issues once per request


hannahtuttle (approved):

netlify/functions/manage-findings.ts
```diff
@@ -38,37 +38,29 @@ import { apiKey, domain } from "../_config";
 
 async function getFinding(
   username: string,
-  contest: Contest,
+  repoName: string,
```
natevw, line 41: so I was going to complain that we lost a more semantic type, but frankly the end result here is **more** readable! 🤔

netlify/functions/manage-findings.ts
```diff
@@ -38,37 +38,29 @@ import { apiKey, domain } from "../_config";
...
 
-  const submission_files = (
-    await getAvailableFindings(client, username, contest)
+  const allFindings = await getAllFindings(
+    client,
+    repoName,
+    process.env.GITHUB_CONTEST_REPO_OWNER!
+  );
+
+  // todo: don't rely on full listing
```
natevw, line 52: isn't this `todo` now done?

netlify/functions/manage-findings.ts
```diff
@@ -135,22 +94,36 @@ async function getFindings(
...
+  );
+
+  const wardenAndTeamFindingRequests: Promise<void | void[]>[] = [];
   if (includeTeams) {
     const teamNames = await getUserTeams(username);
-    promises.push(
-      getTeamsFindings(teamNames, octokit, contest, wardenFindingsForContest)
-    );
+    for (const team of teamNames) {
+      wardenAndTeamFindingRequests.push(
```
natevw, line 107: if there's a risk of "Promise.all traffic jam" here you could now use https://github.com/code-423n4/api.code4rena.com/blob/main/Utils/utils.ts#L353 as documented https://gist.github.com/natevw/79e04dba0cda6c87e19eb63a6a4986ba

package.json
```diff
@@ -56,6 +57,7 @@
     "https-browserify": "^1.0.0",
     "isomorphic-dompurify": "^0.19.0",
     "jsonwebtoken": "^8.5.1",
+    "jszip": "^3.10.1",
```
natevw, line 60: Is this actually still needed? (/Can you tell I scrolled past some of the bigger files for now!)

natevw (changes requested):
Ok, I think I need to take a better look at this in the morning too — but so far most looks good.

Marking "request change(s)" already though because afaict the jszip dep isn't used anymore?

netlify/functions/manage-findings.ts
captainmangoC4: No, this is actually an old comment from when the logic was originally written. The idea here is that we could just fetch the single issue and data file from GH, but in fact we get all findings, then filter by warden, and then filter by the issue number we want. 

It's certainly inefficient, but this function should only be called if the user navigates directly to `contests/[contest-slug]/submit?issue=[issueNumber]` (which is not common). If they navigate by clicking "edit" from the findings tab on the contest page, the finding data is just passed through to the form via location state.

netlify/functions/manage-findings.ts
captainmangoC4: Yep, fair point. However, it's extremely rare for a warden to be on more than 1 team, and I don't know of any that are on more than 2, so this will be at most 3 promises - at least for now.

package.json
captainmangoC4: Yep, still used in `github-utils`

netlify/functions/manage-findings.ts
captainmangoC4: Yeah, really good point.

captainmangoC4: Added better type declarations

natevw (approved):

PR #9063, opened 2023-04-10
Adding extra info and official dates for bot race qualifier

burgertime-c4:
This:

<details>
<summary>Adds the "bot race qualifier" tag (and the concept of tags) to the qualifier contest. </summary>
Screenshot 2023-04-10 at 4 07 23 PM
</details>

<details>
<summary>Both shows the coming soon text and adds the time zone to it. Using GMT since that's all that's available without adding another special timezone package (I think)</summary>
qualifier coming soon
</details>

<details>
<summary>Fixes the link in the text that shows when a qualifier has ended.</summary>
qualifier closed

</details>

Edited to add:
<details>
<summary>Adding link to repo in case the user starts their journey on this page</summary>
Screenshot 2023-04-10 at 9 41 01 PM

</details>


src/pages/register/bot.tsx
```diff
@@ -8,9 +8,10 @@ import { WardenFieldOption } from "../../components/reporter/widgets/WardenField
 import { format, isAfter, isBefore } from "date-fns";
 import ProtectedSection from "../../components/ProtectedSection";
 
-// @todo: replace with correct start and end times
-const START = new Date("2023-03-16T20:43:00.000Z");
-const END = new Date("2023-03-16T20:44:30.000Z");
+// @todo: replace with real contest data
+const START = new Date("2023-04-12T20:00:00.000Z");
+const END = new Date("2023-04-12T21:00:00.000Z");
+const repo = "https://github.com/code-423n4/2023-04-frankencoin";
```
Simon-Busch, line 14: Are these the real contest data now ? 

If so, should we remove the comment ?

src/templates/ContestLayout.tsx
```diff
@@ -193,6 +193,19 @@ const ContestLayout = ({ data }) => {
                 <div>
                   <h1 className="type__headline__xs">{title}</h1>
                   <p>{details}</p>
+                  {contestid === 231 && (
```
Simon-Busch, line 196: Should we *in the future* handle this in the database to avoid hardcoding the contestId ?

Simon-Busch (approved):
Just left a couple of comments but it seems good to go for me

src/templates/ContestLayout.tsx
geoffchan23: @Simon-Busch yes! But I believe we can leave this hardcoded for now as this is the first qualifier. Let's make sure to clean this up afterward.

PR #9084, opened 2023-04-11
Register warden 0xDACA

code423n4:
Auto-generated PR to register the new warden 0xDACA

@danielcohenhillel


SamusElderg:
The user is probably `daca#2171` but has not yet checked into the `i-want-to-be-a-warden` channel

SamusElderg (approved):

PR #9182, opened 2023-04-13
Fix rego case validations

### LINKED ISSUE #9180: Duplicate Warden/Team handles are possible
_Originally posted by @SamusElderg in https://github.com/code-423n4/code423n4.com/issues/9170#issuecomment-1506400038_

It is currently possible to end up with duplicate .json handles of the same filename inside the `_data\handles` folder. As long as they have different casing of course, ie:
- `handlename.json`
- `HANDLENAME.json`
- `HandleName.json`
- I guess even `handlename.JSON` ?

So far have only spotted it (see reference above) when there is a team handle (`Plasmablocks.sjon`) and warden handle (`plasmablocks.json`) of the same name but different case sensitivity, there might be sanity checks already in place that cater to case sensitivity for warden applications and/or team applications, but seems pretty clear there must not be between warden<>team

Possible solutions:
1. Make sure Warden & Team handles are stored in different folders instead of them all being in `_data\handles\`. *and* make sure there are sanity checks that `.toLowercase()` or similar for each of the Warden applications && team applications.

OR

2. Make sure there are sanity checks that `.toLowercase()` or similar for each of the Warden applications && team applications. *and* also cross reference between team<>warden applications (so make sure there are no duplicate generic handles, not warden or team specific)

Should also maybe toLowercase() standardise all handle names before they are stored, so all .json handle filenames will be like:
- `handlename.json`
- instead of `HandleNAME.json` or any other combination

In the meantime:
- Ill run through and see if this is the only example that exists (fingers crossed)

SamusElderg:
Adds in case standardization with checking if handles exist when creating a team or bot handle. logic is similar to the existing validation for new wardens.

closes #9180 

I cant run a local copy of the repo unfortunately, this will require some extra testing/checking by the team but the changed logic is pretty simple.

For this issue to be completely closed off and ensure windows contributors dont have headaches, `Plasmablocks.json` should be changed to a different filename && handle which will require some coordination between the C4 team and the members of that handle.


captainmangoC4 (approved):

captainmangoC4:
Thanks @SamusElderg!

captainmangoC4:
Tested creating team "captainmango" in the preview env and confirmed the validation error is caught.

PR #9198, opened 2023-04-13
Distribute GitHub API calls among tokens

### LINKED ISSUE #9197: Distribute api calls

captainmangoC4:
closes #9197

### DIFF
--- a/gatsby-node.esm.js
+++ b/gatsby-node.esm.js
@@ -7,7 +7,6 @@ import path from "path";
 import webpack from "webpack";
 import SchemaCustomization from "./schema";
 import { getApiContestData } from "./netlify/util/getContestsData";
-const { token } = require("./netlify/_config");
 
 const privateContestMessage = dedent`
 ## Contest details are not available. Why not?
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -111,7 +111,7 @@
     "build": "gatsby build",
     "develop": "gatsby develop",
     "format": "prettier --write \"**/*.{js,jsx,ts,tsx,json,css,scss}\"",
-    "start": "netlify dev",
+    "start": "netlify dev -p 3000",
```
natevw, line 114: was this committed intentionally?

gatsby-node.esm.js
```diff
@@ -7,7 +7,6 @@ import path from "path";
 import webpack from "webpack";
 import SchemaCustomization from "./schema";
 import { getApiContestData } from "./netlify/util/getContestsData";
-const { token } = require("./netlify/_config");
```
natevw, line 10: it might be nice to fall back to the original token if the new one isn't set (thinking mostly of devs having to update all their own `.env` files now too) but not a blocker for me

natevw (approved):

gatsby-node.esm.js
captainmangoC4: Fair point

PR #9229, opened 2023-04-15
Register warden mahsuni2915

code423n4:
Auto-generated PR to register the new warden mahsuni2915

@mahsuni2915


SamusElderg:
The user exists but has not yet checked into the `i-want-to-be-a-warden` channel, closing PR for now

PR #9244, opened 2023-04-16
Register warden Dark_Hunter

code423n4:
Auto-generated PR to register the new warden Dark_Hunter

@Lakiya7


SamusElderg:
The user is probably `LaKiya_S#5857` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #9247, opened 2023-04-16
Delete team Illini_Blockhain

code423n4:
This auto-generated PR removes the team Illini_Blockhain


SamusElderg (approved):
Looks fine, just deleting team in place of new team name and additional members, same payout address

PR #9279, opened 2023-04-18
Do not POST contests data back to leaderboard function

### LINKED ISSUE #9276: Leaderboard fetch posts unnecessary data back to server
Looks like The calls to fetch the leaderboard data are sending the full graph of contests data **back** to the server:

Screenshot of browser debugging view showing semi-collapsed array of contests data being part of the request

natevw:
should fix #9276


Simon-Busch (approved):
Seems good to me :)

PR #9342, opened 2023-04-20
Fetch findings data via two split requests

natevw:
Depends on https://github.com/code-423n4/api.code4rena.com/pull/613 changes deployed first!


Simon-Busch (approved):
Didn't test locally but it looks good to me

hannahtuttle (approved):
This is working great and it actually made the leaderboard a little faster (still fairly slow in dev environment).

PR #9425, opened 2023-04-24
Register warden nynin9

code423n4:
Auto-generated PR to register the new warden nynin9

@nynin9


SamusElderg:
The user is probably `nynin#7349` but has not yet checked into the `i-want-to-be-a-warden` channel with a matching nickname

SamusElderg:

PR #9504, opened 2023-04-27
Consolidate bot race configuration

natevw:
This cleans up the CA process so that only one file needs to be updated [albeit still in both repos] instead of in two different locations in this one repo.

(This aligns Gatsby with recent API-side config techniques too, see <https://github.com/code-423n4/api.code4rena.com/pull/657>.)


captainmangoC4 (approved):
Tested and it works as expected

natevw:
Thanks, I don't think there's a **huge** rush here though so I'll leave it for an early-week deploy I suppose, just in case?

natevw:
Updated #9600 and the "Bot race walkthrough notes" with these changes.

PR #9601, opened 2023-05-02
Register warden carlomigueldy

code423n4:
Auto-generated PR to register the new warden carlomigueldy

@carlomigueldy


SamusElderg:
The user is probably ` carlomigueldy.eth#1262` but has not yet checked into the `i-want-to-be-a-warden` channel

carlomigueldy:
> The user is probably ` carlomigueldy.eth#1262` but has not yet checked into the `i-want-to-be-a-warden` channel

done just checked in just now

sksb2180 (approved):

PR #9694, opened 2023-05-05
Fix loading issue when report URL is loaded with capital case

geoffchan23:
(this is a proposed fix for https://github.com/code-423n4/buttons/issues/403 iiuc? ^nvw)


captainmangoC4 (commented):
This works, though you'd also need to update the link on the contest page. However, I think this is actually more of a data issue and should be fixed upstream.

captainmangoC4:
Accompanying PR (makes sure report gen makes contest slug lowercase): https://github.com/code-423n4/api.code4rena.com/pull/729

natevw:
https://github.com/code-423n4/buttons/issues/403#issuecomment-1539091243 is the fix here, see https://discord.com/channels/810916927919620096/1104065708507279470/1105243852262088704

PR #9789, opened 2023-05-09
Hide bot race status bar on non-public contests

captainmangoC4:
[should resolve <https://github.com/code-423n4/dotcom-issues/issues/63>]


natevw (approved):
didn't set up a full test repro but mirrors some existing code in ContestTile and looks good to me

PR #9935, opened 2023-05-18
Register warden Greenstar

code423n4:
Auto-generated PR to register the new warden Greenstar

@github.com/sarangsohail


SamusElderg:
The user is probably `Sarang#6854` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #10022, opened 2023-05-20
Register warden snakeys

code423n4:
Auto-generated PR to register the new warden snakeys

@Chamishka-Ishani


SamusElderg:
The user is probably `Chamishka_Ishani#6676` but has not yet checked into the `i-want-to-be-a-warden` channel, closing PR for now

PR #10060, opened 2023-05-22
Mitigation review form enhancements

captainmangoC4:
Improved UX and feature completeness for mitigation review submissions


netlify/functions/manage-findings.ts
```diff
@@ -403,47 +405,69 @@ async function editFinding(
...
+    let labelNameToAdd = "";
+    if (
+      data.mitigationStatus?.newValue === MitigationStatus.MitigationConfirmed
+    ) {
+      labelNameToAdd = "mitigation-confirmed";
+    } else if (
+      data.mitigationStatus?.newValue === MitigationStatus.Unmitigated
+    ) {
+      labelNameToAdd = "unmitigated";
+    }
```
hannahtuttle, line 450: could you add this logic in the lines above here (410-419) where you are specifying what label to remove you could also add the label to add this this is a type of toggle just to make this a little more concise and readable? Not hindering anything but just a suggestion.

src/components/reporter/MitigationReviewSubmissionForm.tsx
```diff
@@ -0,0 +1,348 @@
...
+      )}
+      {state.mitigationStatus === MitigationStatus.New && (
+        <Input
+          name="title"
+          value={state.title}
+          required={true}
+          handleChange={handleChange}
+          label="Title"
+          helpText="Summarize your findings for the bug or vulnerability. (This will be the issue title)"
+          forceValidation={isSubmitted}
```
hannahtuttle, line 292: Is this better than using just validator on the input? I'm just curious about the reason to use forceValidation over validator?

hannahtuttle (approved):

hannahtuttle:
@captainmangoC4 I went through all these changes and they all good, I did make one small suggestion but It's not a hinderance to merging this with main. So I went ahead and approve the pr.

src/components/reporter/MitigationReviewSubmissionForm.tsx
captainmangoC4: We actually use both, technically. The validator function, if passed in, gets called when the input loses focus. We don't need to pass in a custom validator function when the only validation check is for the value to not be empty (in that case, if we pass required, then the input automatically does a validation check when the element loses focus and shows an error if the field is empty). Where `forceValidation` comes in, is when the user tries to submit without ever clicking into the input. If that happens, we need a way to force the input component to do a validation check so that the user isn't just confused about why the form won't submit.

captainmangoC4: Great question though! I should definitely document that somewhere, because I totally see how that would be confusing.

PR #10077, opened 2023-05-23
Register warden luanusm

code423n4:
Auto-generated PR to register the new warden luanusm


SamusElderg:
The user is probably `Luannm#8291` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #10172, opened 2023-05-26
Register warden biltima3

code423n4:
Auto-generated PR to register the new warden biltima3


SamusElderg:
The user is probably `Biltima31#9725` but has not yet checked into the `i-want-to-be-a-warden` channel

SamusElderg:

PR #10246, opened 2023-05-29
Register warden tuanhieukd

code423n4:
Auto-generated PR to register the new warden tuanhieukd


SamusElderg:
The user is probably `nguyentuanhieu#9990` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #10280, opened 2023-05-30
Register warden londonk

code423n4:
Auto-generated PR to register the new warden londonk


SamusElderg:
The user might be `londonk66| Suiswap#9746` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #10293, opened 2023-05-30
Register warden ade1902

code423n4:
Auto-generated PR to register the new warden ade1902


SamusElderg:
The user might be `ade190219#7346` but has not yet checked into the `i-want-to-be-a-warden` channel

SamusElderg:

PR #10394, opened 2023-06-02
Register warden bobbizi

code423n4:
Auto-generated PR to register the new warden bobbizi


sksb2180:
Warden present but didn't checked required channel.

PR #10436, opened 2023-06-04
Register warden omaralirajbd

code423n4:
Auto-generated PR to register the new warden omaralirajbd

@omaralirajbd


SamusElderg:
The user is probably `omaralirajbd#7049` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #10443, opened 2023-06-05
Register warden teerth17

code423n4:
Auto-generated PR to register the new warden teerth17

@teerth17


SamusElderg:
The user is probably `Teerth#6829` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #10458, opened 2023-06-05
Register warden AlexN

code423n4:
Auto-generated PR to register the new warden AlexN


SamusElderg:
The user is probably `alexneg#1314` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #10461, opened 2023-06-05
Register warden Louisforbes147

code423n4:
Auto-generated PR to register the new warden Louisforbes147


SamusElderg:
The user is probably ` Louisforbes#1733` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #10468, opened 2023-06-05
Register warden DANHSANH

code423n4:
Auto-generated PR to register the new warden DANHSANH

@DANHSANH


SamusElderg:
The user is probanly `DANH SANH#0834` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #10484, opened 2023-06-05
Register warden Azamat

code423n4:
Auto-generated PR to register the new warden Azamat


SamusElderg:
The user is probably `azamat1985#5915` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #10511, opened 2023-06-05
Register warden alizai7867

code423n4:
Auto-generated PR to register the new warden alizai7867


SamusElderg:
The user is probably `Alizai786NemoSwap IDO on 06/01#6980` but has not yet checked into the `i-want-to-be-a-warden` channel

SamusElderg:

PR #10525, opened 2023-06-06
Register warden mambanick

code423n4:
Auto-generated PR to register the new warden mambanick

@nicksinch


SamusElderg:
The user might be `nicksinch#3684` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #10527, opened 2023-06-06
Register warden Ali12

code423n4:
Auto-generated PR to register the new warden Ali12


SamusElderg:
The user exists might be `Ali12 | Suiswap IDO on 06/03#8346` has not yet checked into the `i-want-to-be-a-warden` channel

SamusElderg:

PR #10546, opened 2023-06-06
Register warden samcab

code423n4:
Auto-generated PR to register the new warden samcab

@samcxb


SamusElderg:
The user is probably `samcxb | Suiswap IDO 06/03#3683` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #10581, opened 2023-06-06
Register warden ulung

code423n4:
Auto-generated PR to register the new warden ulung


SamusElderg:
The user exists is probably `ulung.id#7003` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #10671, opened 2023-06-08
Register warden trunghieu

code423n4:
Auto-generated PR to register the new warden trunghieu


SamusElderg:
The user is probably `trunghieu2209#4160` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #10689, opened 2023-06-08
Register warden Nextree

code423n4:
Auto-generated PR to register the new warden Nextree

@Pere


SamusElderg:
The user is maybe `pere#7653` but has not yet checked into the `i-want-to-be-a-warden` channel

Issue #11487, opened 2023-07-17
[BUG]: Warden Application | Username hint-text out of date


SamusElderg:
## Describe the bug
The 'Discord Username' field for Warden registrations has note/warning text that is now outdated due to recent Discord changes.
It current says:
`It should end with '#' followed by 4 digits` but this is no longer the case for Discord.

There may also be some validation logic that will also need changing but i have not yet jumped into the code and checked

## Screenshots

![image](https://github.com/code-423n4/code423n4.com/assets/26967164/ebc2c508-595b-42c5-9b78-cc76bae90694)

SamusElderg:
This was solved at some point, closing it out, thankyou 🙏

PR #11501, opened 2023-07-17
fix: #11487 - Warden Application | Username hint-text out of date

raxhvl:
- [x] Updated discord username regex
- [x] Updated discord username hint message

[Discord username help](https://support.discord.com/hc/en-us/articles/12620128861463)

I have forced lowercase username as per documentation. Let me know if we need to allow uppercase as well.


burgertime-c4:
This has a thumbs up from me for content, but I think someone other than me should read the regex 😂

PR #11510, opened 2023-07-18
Update submit-finding.ts to send issueId with submission email

teryanarmen:


SamusElderg:
Thanks so much for this PR, as this is not the repo for the main site anymore ill close this off now, but there is an ongoing process to improve the registration process which will cover the intent of this PR too  🙏

PR #11680, opened 2023-07-24
Register warden amjadali

code423n4:
Auto-generated PR to register the new warden amjadali

@amjadali.110


sksb2180:
Warden didn't checked out the channel `i-want-to-be-a-warden`.

PR #11803, opened 2023-07-30
Register warden flakon34

code423n4:
Auto-generated PR to register the new warden flakon34


SamusElderg:
The user might be `flakon3724` but has not yet checked into the `i-want-to-be-a-warden` channel

SamusElderg (approved):

PR #11872, opened 2023-08-02
Register warden 7230xmurali

code423n4:
Auto-generated PR to register the new warden 7230xmurali

@https://github.com/Potlimuralimohan


SamusElderg:
The user might be `potlimuralimohan` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #12260, opened 2023-08-21
Register bot Bolo

code423n4:
Registration for bot Bolo submitted by PNS.

Description:
The bot analyzes syntax using a lexer and parser with the help of ANTLR. An interesting fact is that the entire system is written in PHP 8.2, which is likely a rarity =)

@pronobis4


thebrittfactor:

PR #12272, opened 2023-08-21
Register bot sstan

code423n4:
Registration for bot sstan submitted by Osiris.

Description:
`sstan` is a static analysis bot that uses ast analysis to identify patterns. We use `solang-parser` and a closed source analyzer to generate findings. Some patterns were inspired by `solstat`, but the core architecture is completely different than the approach taken by `solstat`

@0xOsiris


thebrittfactor:

PR #12373, opened 2023-08-28
Delete team CyberPunks

code423n4:
This auto-generated PR removes the team CyberPunks


itsmetechjay:
Since the team has at least one submitted finding, we won't delete the team. Will create a separate PR to delete members.

PR #12526, opened 2023-09-05
Register warden 0xnote

code423n4:
Auto-generated PR to register the new warden 0xnote

@note89


sksb2180:
Warden didn't checked out the channel `i-want-to-be-a-warden` yet.

SamusElderg:
The user might be `Nils Eriksson` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #12647, opened 2023-09-10
Register warden kongtaoxing

code423n4:
Auto-generated PR to register the new warden kongtaoxing

@kongtaoxing


SamusElderg:

kongtaoxing:
> The user exists but has not yet checked into the `i-want-to-be-a-warden` channel

I've checked in now. Thank you.

sksb2180 (approved):

PR #12744, opened 2023-09-13
Register warden sv1412

code423n4:
Auto-generated PR to register the new warden sv1412

@letonchanh


sksb2180:
Warden didn't checked out the channel `i-want-to-be-a-warden` yet.

SamusElderg:
The user might be `letonchanh#2068` but has not yet checked into the `i-want-to-be-a-warden` channel

PR #13114, opened 2023-10-03
Register warden Adamgue

code423n4:
Auto-generated PR to register the new warden Adamgue


sksb2180:
The user has not yet checked into the `i-want-to-be-a-warden` channel

PR #13453, opened 2023-10-16
Register bot Cygnet

code423n4:
Registration for bot Cygnet submitted by rjs.

Description:
*Cygnet* is a completely new Solidity static analysis tool written from scratch in F#.

Major features:
- A custom `solc` compiler front-end that can handle Truffle, Hardhat, Forge, or nearly any other Solidity project system. The benefit of this approach is that *Cygnet* can take advantage of the advanced output selection modes present in `solc` that are not enabled by default. This additional output allows *Cygnet* to perform more **extensive** and **accurate** static analysis than other static analysis tools which rely on regex or AST analysis, for example, the `solc` struct layout output may be utilized even if not enabled by the project.

- Backwards compatible `solc` output loaders going back to `solc` version 5. This allows *Cygnet* to safely skip over unsupported fields and features in older compilers, while producing more extensive analysis reports on newer compiler outputs.

- F# static typing of entire `solc` AST ensures *Cygnet* is extremely accurate, fast and stable. Unlike other bots which rely mostly on regex expressions, you will note that *Cygnet* often picks up issues on e.g. multiple lines whereas the other bots fail due to their regex-based architecture. Development of new analyzers can leverage an extensive library of `solc` extensions to produce new analyzers within minutes.

- `MermaidJS` diagrams of e.g. recommended function order in classes -- unfortunately work on this has been suspended/disabled due to the proposed move to `JSON` submissions.

- Future plans include CVE integration, Slither integration, GPT-4 based notes using a historical report-based fine-tuned model (if it works!)

@ryanjshaw

### DIFF
--- a/_data/bots/Cygnet.json
+++ b/_data/bots/Cygnet.json
@@ -1,0 +1,16 @@
+{
+  "handle": "Cygnet",
+  "crew": [
+    "rjs"
+  ],
+  "paymentAddresses": [
+    {
+      "chain": "polygon",
+      "address": "0x0FDdab9cf8e426d09e58640E5c647897A943F99a"
+    },
+    {
+      "chain": "ethereum",
+      "address": "0x0FDdab9cf8e426d09e58640E5c647897A943F99a"
+    }
+  ]
+}


thebrittfactor:

PR #14584, opened 2023-12-16
Register warden HoangQuoc

code423n4:
Auto-generated PR to register the new warden HoangQuoc

@thailosecurity


SamusElderg:
The user might be `thailosoquit17` but but has not yet checked into the `i-want-to-be-a-warden` channel

SamusElderg:

PR #15633, opened 2024-02-16
recover deleted teams

Carlos-stackfive:
Re-add the teams that were deleted [here](<https://github.com/code-423n4/code423n4.com/pull/14660/files>), [here](<https://github.com/code-423n4/code423n4.com/pull/14651/files>), [here](<https://github.com/code-423n4/code423n4.com/pull/14657/files>), and [here](<https://github.com/code-423n4/code423n4.com/pull/14016/files>)

### DIFF
--- a/_data/handles/Audit_Avengers.json
+++ b/_data/handles/Audit_Avengers.json
@@ -1,0 +1,20 @@
+{
+  "handle": "Audit_Avengers",
+  "link": "https://twitter.com/JP_Courses",
+  "moralisId": "",
+  "showOnLeaderboard": true,
+  "members": [
+    "JP_Courses",
+    "pxng0lin"
+  ],
+  "paymentAddresses": [
+    {
+      "chain": "polygon",
+      "address": "0x82aAE2d95b92dD5576DbC59F455521250bb9396A"
+    },
+    {
+      "chain": "ethereum",
+      "address": "0x82aAE2d95b92dD5576DbC59F455521250bb9396A"
+    }
+  ]
+}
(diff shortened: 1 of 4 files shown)


_data/handles/Audit_Avengers.json
```diff
@@ -0,0 +1,20 @@
...
+  "paymentAddresses": [
+    {
+      "chain": "polygon",
+      "address": "0x82aAE2d95b92dD5576DbC59F455521250bb9396A"
+    },
+    {
+      "chain": "ethereum",
+      "address": "0x82aAE2d95b92dD5576DbC59F455521250bb9396A"
+    }
+  ]
```
captainmangoC4, line 19: Were these teams deleted on purpose? Meaning, did the team members _want_ to retire their team? If so, we should set showOnLeaderboard to false and consider clearing out payment addresses.

natevw (approved):
lgtm especially w/now leaving them `showOnLeaderboard: false`. the `paymentAddresses` have already been public so personally don't see a reason for any further adjustments.

_data/handles/Audit_Avengers.json
liveactionllama: @captainmangoC4 & @Carlos-stackfive - Correct! All 4 of these were teams that requested to have their team deleted. In addition to hiding from the leaderboard, it would be nice if we can clear the members and payment addresses as well (that's part of the standard process that had been discussed with mango quite some time ago).

Carlos-stackfive: Ok, I will clear out the members as well; though I think leaving the payment addresses in for now might be better so as to retain wallet address data across the board, particularly for OFAC compliance.
In the future when stuff begins to move over to v1, we'll have soft deletes.

PR #15677, opened 2024-02-18
Register warden robertod99

code423n4:
Auto-generated PR to register the new warden robertod99

@robertod99


robertod99:
Please close, I created this because the first time I signed up with my other account I wrote my discord user name with the discriminator which is going to be remove soon so I though it could cause some problems. I don't want this to look like I am trying to sign up twice, I will wait for my first sign up to get approved by adminstrators.

PR #16002, opened 2024-03-05
fix broken invite url

ep-c4:


nlf (approved):
confirmed this link is active and working :+1:

PR #16834, opened 2024-04-22
Fix validator for teams, remove the rest

Carlos-stackfive:
After the auth system transition, the validator for teams will not work as expected as this repo no longer acts as a mediator for newly registered users. The consequence is that newly registered users who are listed in a team's `members` field will incorrectly flag an existing user as non-existent.

The main goal of this PR is to replace the teams validator. It should now work as intended, plus improvements to performance, output messages, as well as to additionally output a warning (for team creations) if a handle is already taken. 

This PR also cleans up validators that should no longer be put to use.

Development was done in my own separate repo with actions and repo structure set up identically.

Leverages https://github.com/marketplace/actions/changed-files


bin/validate.js
```diff
@@ -1,206 +1,68 @@
...
-  for (const orgFile of orgs) {
-    const blob = await readFile(orgFile);
-    let parsedOrg;
-    try {
-      parsedOrg = JSON.parse(blob);
-    } catch (err) {
-      console.error(`Unable to parse JSON file at ${orgFile}`);
+    // check that the required members field exists
+    const teamMembers = parsedHandle.members;
+    if (!teamHandle) {
```
natevw, line 32:
```suggestion
    if (!teamMembers) {
```

Carlos-stackfive: Ah great catch

natevw (approved):
Looks like a reasonable approach to me! (Mostly just read the new code as its own whole file and didn't compare to what the old was all doing since the diff view here is a mess and most of the data needs have changed anyway…)

SamusElderg (commented):
@Carlos-stackfive @natevw just pinging you in based on the comments below regarding this validation workflow suiting CA processes (currently)

bin/validate.js
```diff
@@ -1,206 +1,68 @@
...
     }
-
-    if (!Object.prototype.hasOwnProperty.call(parsedOrg, "image")) {
-      console.error(`Unable to find key "image" in ${orgFile}`);
+    // check that there is at least 1 member
+    if (teamMembers.length < 1) {
+      console.error("❌ Team must have 1 or more members");
       passedValidation = false;
       continue;
     }
```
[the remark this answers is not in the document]
natevw: ah, I thought we were perhaps proceeding on soft-delete and/or self-serve for this stuff — but afaict the next round of WMA improvements has been delayed. emptying out the members would indeed be a shortcut and I personally would be fine removing this check (assuming it's not guarding some sensitive API v0 code that gets upset in that case).

PR #17242, opened 2025-09-01
Create team CertiK

code423n4:
This auto-generated PR creates the team CertiK


CloudEllie:
I've asked the wardens forming this team to confirm that they're affiliated with CertiK; otherwise, they should select a different team name for the avoidance of confusion.

